PR #23, opened 2013-03-21
Make pundit methods non-public

cpuguy83:
Pundit methods should not be public as public methods show up as controller actions, which they are not.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -34,11 +34,8 @@ module Pundit
     if respond_to?(:helper_method)
       helper_method :policy_scope
       helper_method :policy
-
-      protected :policy_scope
-      protected :policy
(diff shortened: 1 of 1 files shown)


iamvery:
Sorry to bring an old PR out of the woodworks. TIL `hide_action`, and I'm curious why this is preferable to just marking methods as `private` or `protected. I've always used access modifiers for non-action methods on a controller. Is this bad practice? (thanks and please forgive the question :laughing:)

jnicklas:
There are some edge cases where it matters. For example it's nicer to call these methods in a controller test, or when working with a controller instance in the console. Mostly it's the same though.

iamvery:
Thanks for the info (and patience with me) @jnicklas!

PR #50, opened 2013-08-12
Allow authorize to pass arguments to policy methods.

coop:
`authorize` currently assumes that each policy has enough information to make a decision. This isn't always the case as described in the policy below.

``` ruby
class ProjectPolicy
  # Determine if user is allowed to remove other_user from the project.
  #
  # @return [Boolean]
  def remove_user?(other_user)
    user != other_user && project.owned_by?(user)
  end
end
```

`authorize` has been patched to send any extra arguments to the policy which 
makes the above possible.

``` ruby
class ProjectsController < ApplicationController
  include Pundit

   # This isn't RESTful but shows intent.
  def remove_user
    authorize project, :remove_user?, other_user

    project.remove_user other_user
    respond_with project
  end

   private

  def project
    @project ||= current_user.find_project params[:id]
  end

  def other_user
    @other_user ||= project.find_user params[:user_id]
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,8 @@ Pundit makes the following assumptions about this class:
   you want to check. This does not need to be an ActiveRecord or even
   an ActiveModel object, it can be anything really.
 - The class implements some kind of query method, in this case `create?`.
-  Usually, this will map to the name of a particular controller action.
+  Usually, this will map to the name of a particular controller action. The
+  query method may take zero or more arguments.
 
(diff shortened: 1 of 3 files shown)


jnicklas:
I'm closing this due to the same reason as #133. Please see [my explanation](https://github.com/elabs/pundit/pull/133#issuecomment-86881448) there. Please respect that this is my final decision on the matter.

PR #129, opened 2014-04-04
Unspaghettify PolicyFinder#find

zeevallin:
There is no way nested if-else's can be healthy.


darkleaf:
It's cool PR.
@thomasklemm  https://github.com/bbatsov/ruby-style-guide#no-nested-conditionals

PR #133, opened 2014-04-11
pass arbitrary arguments to #authorize

lanej:
this facilitates the usecase where an query method would want to conditionally authorize based on the content of the request parameters

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -59,12 +59,12 @@ module Pundit
     raise AuthorizationNotPerformedError unless @_policy_scoped
   end
 
-  def authorize(record, query=nil)
+  def authorize(record, query=nil, *args)
     query ||= params[:action].to_s + "?"
     @_policy_authorized = true
 
     policy = policy(record)
-    unless policy.public_send(query)
+    unless policy.public_send(query, *args)
       error = NotAuthorizedError.new("not allowed to #{query} this #{record}")
       error.query, error.record, error.policy = query, record, policy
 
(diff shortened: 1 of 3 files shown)


jnicklas:
So this PR has been open for a long time, and activity here has died down a bit. This is probably the most requested feature for Pundit, and so I realize that this will be a bit of an unpopular move, but I have decided to close this PR without merging it.

Here are my reasons for doing this:

- Currently using policies is straightforward and obvious. The query method dictates what we are asking, and the object dictates what we are asking from. In the cases where multiple objects are involved in whether a particular action should be allowed, they should be wrapped up in something, for example by utilizing the form object pattern. This is good practice anyway, and Pundit is all about encouraging good practice.

- Adding parameters to the query methods makes them harder to use in views, adding parameters to the policy constructor makes the caching situation of policies (introduced after this PR iirc) much more non-obvious.

- The usecase that @lanej has described, which I think is valid, is much less relevant given the establishment of the `permitted_attributes` pattern. I have an upcoming PR which expands this functionality and should solve @lanej's usecase in a much more elegant way.

- The context pattern established in the README since this PR solves a lot of what people want to use this kind of pattern for, and it's a much better, much more convenient solution. We use this very successfully in a multi-tenant app, and it works very nicely. If we would have to pass in the tenant everywhere, a lot of the convenience of Pundit would be lost.

- This is a feature which is very, very easily abused, and by Murphy's law **will** be abused. The original intent behind Pundit was establishing a simple good-practice pattern for authorization, we should not be encouraging bad patterns.

I want to clarify that I have thought about this a lot. Please respect that this is my final decision on the matter.

lanej:
Proper explanation. Thanks for the time @jnicklas

PR #136, opened 2014-04-17
add stand-alone authorizer

billychan:
I found a stand alone authorizer is desired sometimes, when there is no involvement of controller.

I've wrote such class in my app's lib and being quite happy to use it for days. I guess the main repo may need this as well to provide more support to service object.

Test and doc can be added later if the idea is okay.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -1,4 +1,5 @@
 require "pundit/version"
+require "pundit_authorizer"
 require "pundit/policy_finder"
 require "active_support/concern"
 require "active_support/core_ext/string/inflections"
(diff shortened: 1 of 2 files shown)


openface:
@paulccarey Any progress to report on your findings here?  We're too looking to integrate Mutations gem (service objects) with a simple Authorization routine.   I'm keen on using Pundit, but we have a mixed case of needing to perform authorization from both controllers and from the service objects, so a stand-alone authorization solution would be ideal.

paulccarey:
@openface this slipped my mind if I'm honest due to other pressing priorities, however thanks for the reminder, I'll try to find some time over the coming days and put it on my todo list as this could become a real issue for us in the near future.

openface:
@paulccarey Does #227 address this fully?

Issue #130, opened 2014-04-05
RSpec + Pundit; Undefined method all? and none?


tkeifer:
This may be something really simple, but I was unable to find any answer after extensive googling. 

I'm trying to run the following RSpec test:

```
require "spec_helper"

describe ReportPolicy do
  subject { ReportPolicy.new(user, report) }

  let(:user)    { FactoryGirl.create(:user) }
  let(:report)  { FactoryGirl.create(:report, organization_id: user.organization_id) }

  context "for a user" do
    it { should       permit(:show) }
    it { should_not   permit(:edit) }
  end  
end
```

It is failing with the errors:

```
1) ReportPolicy for a user should permit :show
     Failure/Error: it { should       permit(:show) }
     NoMethodError:
       undefined method `all?' for nil:NilClass
     # ./spec/policies/report_policy_spec.rb:10:in `block (3 levels) in <top (required)>'
2) ReportPolicy for a user should not permit :edit
     Failure/Error: it { should_not   permit(:edit) }
     NoMethodError:
       undefined method `none?' for nil:NilClass
     # ./spec/policies/report_policy_spec.rb:11:in `block (3 levels) in <top (required)>'
```

Can anyone explain what is going on here?

ColinW520:
thanks for this thread - just saved me a ton of time

tirdadc:
I wish I had found this before I ended up renaming the custom matcher to fix the issue. Thanks for the `policy_scope` spec example!

chickenfingers:
failure_message_for_should and failure_message_for_should_not are deprecated. Use failure_message and failure_message_when_negated instead.

PR #137, opened 2014-04-17
add rspec example for scope on an index collection

netzfisch:
Hi @thomasklemm,
as discussed before in issue #130 here the PR, which

* adds a rspec example for authorising single records of an index collection
* adds rspec hints concerning custom matcher when scoping to an user context

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,10 +435,57 @@ describe PostPolicy do
   end
 end
 ```
+An alternative approach to Pundit policy specs is scoping them to a user context
+as outlined in this [excellent post](http://thunderboltlabs.com/blog/2013/03/27/testing-pundit-policies-with-rspec/).
+In that case be aware that you need
 
-An alternative approach to Pundit policy specs is scoping them to a user context as outlined in this
-[excellent post](http://thunderboltlabs.com/blog/2013/03/27/testing-pundit-policies-with-rspec/).
+- to enable a custom matcher as described in this related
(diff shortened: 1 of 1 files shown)


tirdadc:
When trying this approach, I get this error:

```
Failure/Error: expect(policy_scope).to eq [item]
       expected: [#<Item id: 14, name: "Item 16", created_at: "2015-02-02 13:56:48", updated_at: "2015-02-02 13:56:48">]
            got: #<ActiveRecord::Relation []>
       (compared using ==)
```
`policy_scope` is returning an `ActiveRecord::Relation`, not an array of ActiveRecords.

netzfisch:
Hi @thomasklemm,
is there any issue why you didn't consider my pull-request?

Thanks for feedback.

jnicklas:
I'm :-1: on this PR. We should work to keep the README short. It makes sense to explain the `permit` matcher for policies, since that is provided by Pundit.

IMO any competent Rails developer with RSpec experience should be able to write this kind of spec, there's really nothing notable about it. We instantiate the object under test, and do a couple of assertions. I don't know why that would merit explanation.

richeterre:
For what it's worth, I found the scope example really useful, in particular the

```ruby
subject(:policy_scope) { PostPolicy::Scope.new(user, scope).resolve }
```

line. As a Pundit newbie, it would've taken me quite some time to figure it out. Thanks @netzfisch!

Issue #77, opened 2013-11-17
Authorizing controller actions with no associated model


pbougie:
Pundit looks great for authorizing models but in some instances I need to authorize controller actions (with no associated model) as well. A simple check whether the user is logged in or not will suffice most of the time. However, sometimes the ability to view a controller action is based on a user role, i.e. user can view settings page if administrator but not if regular user.

What are the best practices for these situations? Any clarifications would be appreciated on how all the authorization pieces fit together.

Senich:
Is this feature is available in current gem version? Seems for me it doesn't work properly.
I've still got an error  - "unable to find policy `DashboardPolicy` for `:dashboard`" while I do have dashboard_policy.rb in policies folder.

thomasklemm:
Yes, it is. Maybe you'd need to restart your Rails server to pick up the changes or have a typo either in the file name or class name. Could you check that?

Senich:
I double checked that, there is no typo and of course I restarted server couple times. It's so simple so I have no idea where I screwed up.
**dashboard_controller.rb**

     class DashboardController < ApplicationController
        def index
          authorize :dashboard, :index?
          # skip_policy_scope
        end
      end
**dashboard_policy.rb**

    class DashboardPolicy < Struct.new(:user, :dashboard)
      class Scope < Scope
        def resolve
          scope
        end
      end
  
      def index?
        true
      end
  
    end 
my root path is 'dashboard#index' so when I start my application I get beautiful 
Pundit::NotDefinedError in DashboardController#index
unable to find policy `DashboardPolicy` for `:dashboard`   and again policy is in the policies/dashboard_policy.rb

nbekirov:
Is it `app/policies/dashboard_policy.rb` or `policies/dashboard_policy.rb`?

---

Check this out:
  * If I make a typo in my policy class name like so

  ```ruby
  class DashboardPolicyWrong < Struct.new(:user, :dashboard)
  ```
  The error is as follows: `Unable to autoload constant DashboardPolicy, expected [APP_DIR]/app/policies/dashboard_policy.rb to define it`

  * However if the file is missing (if I move/rename/delete it) the error is `unable to find policy DashboardPolicy for :dashboard`

Senich:
sorry, policy is in the app/policies/ of course
Your example is correct providing pundit can initially see your policy but in my case pundit somehow can't locate that

Senich:
Interesting, in this case I get
[...]

PR #141, opened 2014-04-28
Add policy_attributes helper

thomasklemm:
This PR adds a `policy_attributes` helper. It wraps the current approach (of encouraging `PostPolicy#permitted_attributes` to return accessible attributes).

```ruby
def post_params
  params.require(:post).permit(*policy_attributes(@post || :post))
end
```

If you don't need the model to decide which attributes are permitted, it gets even shorter:
```ruby
def post_params
  params.require(:post).permit(*policy_attributes(:post))
end
```

I've tried wrapping things even further and returning some kind of strong params object, however this seems to be cleanest and nicest approach without bloating Pundit too much.

One "side effect" is that you can also pass policies as symbols, which allows policies without underlying models as suggested in #77: `authorize :dashboard, :show? # => DashboardPolicy#show?`

@jnicklas I'd highly value your opinion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,12 +369,14 @@ In Rails 4 (or Rails 3.2 with the
 [strong_parameters](https://github.com/rails/strong_parameters) gem),
 mass-assignment protection is handled in the controller.
 Pundit helps you permit different users to set different attributes. Don't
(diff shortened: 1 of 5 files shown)


jnicklas:
Superseded by #261

Issue #12, opened 2013-01-19
Namespaced policies


assembler:
The app I'm working on has two sections: user and admin. There are multiple types of users and multiple types of admins, and different authentication solutions are used for each of them.

Although all the application models are used by both admins and users, policies are very different. Handling both users and admins and all distinct types within single policy was not so clear and full of `if`s. 

The way I handled this was to dynamically define `policy_class` on every `ActiveRecord` descendant in before filter. Something like this:

``` ruby
def set_policy_namespace
  m = self.class.name.split("::").first
  active_record_descendants.each do |ar|
    ar.class.instance_eval do
      define_method(:policy_class) do
        "::Policies::#{m}::#{self.name}Policy"
      end
    end
  end
end

def active_record_descendants
  Rails.application.eager_load! unless Rails.application.config.cache_classes
  ActiveRecord::Base.descendants
end
```

So, if I'm in `Admin::HomeController`, `Project` model will have `policy_class` redefined to return `::Policies::Admin::ProjectPolicy`. 

I've tried to demonstrate valid use case for this. Do you have any thoughts on how can this be made easier? I'm not sure whether or not it should be a part of gem (since it is somewhat rare).. I'm just asking for opinion.

thanks

coryodaniel:
I just released a gem the other day. It is compatible with Pundit's policies and allows you to namespace based on controller, model or both. Check it out if you still need this functionality. [Regulator](https://github.com/coryodaniel/regulator)

andreimoment:
Thank you!

Hope you can add "Use" section with a few examples to the readme... The
tests don't speak clearly to me.

coryodaniel:
Will do this Sunday. No problem!

andreimoment:
And -- I appreciate your releasing. It. Model-based permissions feel
strange for me as most of my apps have Admin-namespaced controllers. I have
an app ready to use this !

martinverdejo:
how do I implement the `policy:` option from [here](https://github.com/elabs/pundit/issues/12#issuecomment-33827510)?

I have this working:
```raise Pundit::NotAuthorizedError unless Admin::HomePolicy.new(current_user, :home).index?```

but not this:
```authorize :home, :index?, policy: Admin::HomePolicy```

is this still the best approach?

TangMonk:
Still do not know how to use, where is the documentation?

thomasklemm:
@TangMonk Namespaced policies are not supported right now.

TangMonk:
[...]

PR #152, opened 2014-05-22
Lookup policies in the current namespace

ecbypi:
Addresses #12. If the policy is not defined in `namespace`, `const_get`
will search through the inheritance change of `namespace` to find the
policy.


llekn:
Hi guys!

I saw this PR but I don't see the merged code anymore in the `master` branch.

I think it is a really useful feature... Why it was removed?

PR #168, opened 2014-07-13
Enable headless policies

thomasklemm:
Enables **`policy(:dashboard) # => DashboardPolicy`**. 

Policies without a matching model can come in handy when a controller
isn't modeled alongside a resource, e.g. a `DashboardsController`.

The policy lookup by symbol also helps with strong parameters,
since I prefer `policy(:post)` or `policy(@post || :post)` over `policy(@post || Post)`.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -55,6 +55,8 @@ class ArticleTag
   end
 end
 
+class DashboardPolicy < Struct.new(:user, :dashboard); end
+
 class Controller
   include Pundit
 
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -55,6 +55,8 @@ def destroy?
   end
 end
 
+class DashboardPolicy < Struct.new(:user, :dashboard); end
```
Jainish: What is this for?

PR #170, opened 2014-07-15
Find correct policy when model is namespaced

Unknown-Guy:
Fixes the issue with namespaced models https://github.com/elabs/pundit/pull/152#discussion_r14923599

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -15,7 +15,7 @@ module Pundit
 
     def policy
       klass = find
-      klass = namespace.const_get(klass) if klass.is_a?(String)
+      klass = namespace.const_get(klass.demodulize) if klass.is_a?(String)
       klass
     rescue NameError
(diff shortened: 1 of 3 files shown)


coryodaniel:
I've been googling around looking for this functionality in Pundit. Curious if it will ever be merged into master / release gem. I have something very similar hacked in for creating different policies for activeadmin vs the user facing version. Just curious when I can ditch the monkey patching I have in place.

coryodaniel:
I've been looking for this functionality and noticed that it looks to be implemented in a different way. Just wanted to point it out for anyone else who was following along in this PR or in https://github.com/elabs/pundit/pull/152#discussion_r14923599.

The relevant commit showing how to do namespacing for something like having policies for ActiveAdmin and a separate set for your user facing site: https://github.com/elabs/pundit/commit/9711b4896b175fbebec3d5d090821d9ce5fc6728

PR #206, opened 2014-09-05
change policy finder find order

bonyiii:
I worked in a legacy app and sadly it has a model which has `model_name` as an attribute. I find it less likely that a class `model_name` is overwritten, that's why I propose this change in the find order.

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -36,10 +36,11 @@ module Pundit
       elsif object.class.respond_to?(:policy_class)
         object.class.policy_class
       else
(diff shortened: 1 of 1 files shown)


thomasklemm:
I think the intention of asking the `object.respond_to?(:model_name)` first is that people can override the `model_name` on the class, with is tested later on with `object.class.respond_to?(:model_name). Sadly, with this change, it would not be possible to override the `#model_name` on an object any more and the class-level `.model_name` would always take preference. That's not what people would expect, thus I think we should keep the current lookup chain. Please patch pundit yourself or rename the `model_name` column.

PR #216, opened 2014-10-11
Add array-based syntax for namespaced headless policies

RaZer:
Before to namespace headless policy we needed to use following syntax:
```
authorize :'project/dashboard'
```

Now we can use array to specify each constant separately so it looks cleaner
```
authorize [:project, :dashboard]
```

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -44,6 +44,8 @@ module Pundit
           object
         elsif object.is_a?(Symbol)
           object.to_s.classify
+        elsif object.is_a?(Array)
+          object.map(&:to_s).join('/').to_s.classify
         else
           object.class
         end
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -102,6 +102,13 @@ describe Pundit do
         expect(policy.user).to eq user
         expect(policy.dashboard).to eq :dashboard
       end
+
+      it "returns an instantiated policy given an array" do
+        policy = Pundit.policy(user, [:project, :dashboard])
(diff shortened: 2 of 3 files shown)


thomasklemm:
Looks good to me, I'm :+1: on this addition. It's the same as in Rails, where `link_to [:project, :dashboard]` stands for `link_to project_dashboard_path`.

johnotander:
I'd love for this to be included at some point.

lib/pundit/policy_finder.rb
```diff
@@ -44,6 +44,8 @@ def find
           object
         elsif object.is_a?(Symbol)
           object.to_s.classify
+        elsif object.is_a?(Array)
+          object.map(&:to_s).join('/').to_s.classify
```
johnotander: Can't this be put more simply as: `object.join('/')` since `to_s` is implicitly called or am I missing something?

```
2.2.0 :001 > [:a, :b].join('/')
 => "a/b" 
2.2.0 :002 > class Foo
2.2.0 :003?>   end
 => nil 
2.2.0 :004 > [Foo.new, :b].join('/')
 => "#<Foo:0x007f861a1686e8>/b" 
```

jnicklas: Yes. Fixed.

jnicklas:
Merged! Thank you, nice idea :)

jnicklas:
It actually just occurred to me. Can this be extended to allow truly namespaced policies? Something like `policy(:admin, @post)`? It would have to recurse through the array of course, but it might be doable!

emolayi:
Will namespacing be extended to the policy_scope method, @jnicklas?

PR #219, opened 2014-10-20
fixed indentation issue with if statement

jdublu10:
Found from http://refactorcop.r14.railsrumble.com/


thomasklemm:
Thanks for taking a look so closely. However, I think the `end` is indeed indented correctly, it's closing the nested `if` statement.

PR #227, opened 2014-11-07
Make Pundit#authorize a module method.

scottjacobsen:
There are cases where service objects could make use of the authorize
method. Make it a module method so it can be called without mixing in
Pundit.

Have the Pundit#authorize instance method delegate to Pundit::authorize
module method.

Be sure the instance method continues to cache the policy if found.

Closes elabs/pundit#226

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -17,6 +17,18 @@ module Pundit
   extend ActiveSupport::Concern
 
   class << self
+    def authorize(user, record, query, policy = nil)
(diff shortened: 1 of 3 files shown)


jnicklas:
Sorry to take so long on this PR. I've merged this with a small change in that I removed the extra `policy` argument to `Pundit.authorize`. Instead I've simply replicated the code in both places. A little less DRY, but more consistent as an interface.

scottjacobsen:
:+1:

PR #229, opened 2014-11-10
More informative error messages for multi permission checks checks.

timolehto:
So, what I was thinking was something like this. Could maybe revert to the original behaviour when there's only one permission, 'cos then repeating the sole violators name is then futile. What do you think?

### DIFF
--- a/lib/pundit/rspec.rb
+++ b/lib/pundit/rspec.rb
@@ -23,7 +23,7 @@ module Pundit
 
         failure_message_when_negated_proc = lambda do |policy|
           was_were = @violating_permissions.count > 1 ? "were" : "was"
-          "Expected #{policy} to grant #{permissions.to_sentence} on #{record} but #{@violating_permissions.to_sentence} #{was_were} granted"
+          "Expected #{policy} not to grant #{permissions.to_sentence} on #{record} but #{@violating_permissions.to_sentence} #{was_were} granted"
         end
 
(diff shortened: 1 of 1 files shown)


lib/pundit/rspec.rb
```diff
@@ -7,19 +7,23 @@ module Matchers
...
         failure_message_proc = lambda do |policy|
-          "Expected #{policy} to grant #{permissions.to_sentence} on #{record} but it didn't"
+          was_were = @violating_permissions.count > 1 ? "were" : "was"
+          "Expected #{policy} to grant #{permissions.to_sentence} on #{record} but #{@violating_permissions.to_sentence} #{was_were} not granted"
         end
 
         failure_message_when_negated_proc = lambda do |policy|
-          "Expected #{policy} not to grant #{permissions.to_sentence} on #{record} but it did"
+          was_were = @violating_permissions.count > 1 ? "were" : "was"
+          "Expected #{policy} to grant #{permissions.to_sentence} on #{record} but #{@violating_permissions.to_sentence} #{was_were} granted"
```
thomasklemm: Should be `Expected ... **not** to grant ...` in the beginning, right?

thomasklemm:
Thanks for your PR. One small comment, otherwise looks good to me :+1:

jnicklas:
Merged! This is very nice, I've found myself frustrated by this terrible error message many times :)

PR #250, opened 2015-02-20
Add `#skip_authorization` helper

nickrivadeneira:
Allows users to skip authorization and prevent exceptions that result from
having `#verify_authorized` enabled. There are some times when you'd only
conditionally want to skip authorization, but you still want to keep
verification enabled for that action.

An example:
```ruby
def show
  record = Record.find_by(attribute: "value")
  if record.present?
    authorize record
  else
    skip_authorization
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,22 @@ class ApplicationController < ActionController::Base
 end
 ```
 
+If you're using `verify_authorized` in your controllers but need to
+conditionally bypass verification, you can use `skip_authorization`. This is
+useful in circumstances where you don't want to disable verification for the
+entire action, but have some cases where you intend to not authorize.
+
+```ruby
+def show
+  record = Record.find_by(attribute: "value")
+  if record.present?
+    authorize record
+  else
+    skip_authorization
+  end
+end
+```
+
 ## Scopes
 
 Often, you will want to have some kind of view listing records which a
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -77,6 +77,10 @@ module Pundit
     true
   end
 
+  def skip_authorization
+    @_pundit_policy_authorized = true
+  end
+
   def policy_scope(scope)
     @_pundit_policy_scoped = true
     policy_scopes[scope] ||= Pundit.policy_scope!(pundit_user, scope)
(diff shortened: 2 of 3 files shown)


andoq:
+1 for this idea, just ran into this as well.  For now, I've added a skip_authorize? method to my policies that returns true, so I can do:

```
else
  authorize(Model, :skip_authorization?)
end
```

It's not the cleanest API, but at least the intent is relatively clear.

jnicklas:
Seems like a good idea to me! @thomasklemm what do you think?

nickrivadeneira:
I've added documentation to the README so users know it's available.

thomasklemm:
Seems like a useful idea to me, too. :+1: 
Great PR @nickrivadeneira. Thanks for documenting the change in the Readme, too!

README.md
```diff
@@ -177,6 +177,22 @@ class ApplicationController < ActionController::Base
 end
 ```
 
+If you're using `verify_authorized` in your controllers but need to
+conditionally bypass verification, you can use `skip_authorization`. This is
+useful in circumstances where you don't want to disable verification for the
+entire action, but have some cases where you intend to not authorize.
```
iwz: Why not use semantic linefeeds? http://rhodesmill.org/brandon/2012/one-sentence-per-line/

iwz:
:+1:

README.md
nickrivadeneira: I think a change like that would be better in a full documentation change, but for the purposes of this PR I'd like to remain consistent with the existing style.

fantasticfears:
I am confused at `verify_policy_scoped`. Doesn't it be escaped?

nickrivadeneira:
@fantasticfears Can you clarify? I'm not sure what you're asking.

fantasticfears:
Sorry for not being specific. My question is how can I skip `verify_policy_scoped`? `skip_authorization` will not set the `@_pundit_policy_scoped`.

jnicklas:
Merged!

jnicklas:
@fantasticfears we should probably have a separate `skip_policy_scope` method. PRs welcome ;)

hachpai:
For me, I've monkey patched authorize in a way to check if object is nil. Moreover, if not nil I check model validation, to avoid nil fields raising exception with checking operations. I think that this second part would be specially interesting to add, in order to know if all the fields are well suited to compute authorization.

PR #253, opened 2015-02-25
 Don't look for NilClassPolicy

### LINKED ISSUE #251: Don't look for NilClassPolicy
In case `authorize` receives `nil` it should just deny authorization, not look for `NilClassPolicy `. I can't imagine any situation where someone would actually want to implement `NilClassPolicy `.

michaldarda:
Hello, small fix, that @jnicklas talked about

Fixes #251

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -8,7 +8,17 @@ require "active_support/dependencies/autoload"
 
 module Pundit
   class NotAuthorizedError < StandardError
-    attr_accessor :query, :record, :policy
+    attr_reader :query, :record, :policy
+
+    def initialize(options = {})
+      @query  = options[:query]
+      @record = options[:record]
+      @policy = options[:policy]
+
+      message = options.fetch(:message) { "not allowed to #{query} this #{record}" }
+
+      super(message)
+    end
   end
   class AuthorizationNotPerformedError < StandardError; end
   class PolicyScopingNotPerformedError < AuthorizationNotPerformedError; end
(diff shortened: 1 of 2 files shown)


jnicklas:
Maybe this should set `#query` on the error?

michaldarda:
Of course, sounds reasonable ;)

I also introduced little refactoring to avoid duplication

lib/pundit.rb
```diff
@@ -9,6 +9,19 @@
 module Pundit
   class NotAuthorizedError < StandardError
     attr_accessor :query, :record, :policy
+
+    def self.create(options = {})
```
jnicklas: Why not make this a constructor instead?

michaldarda: Right, for some weird reasons I didn't want to override default error constructor. Changed ;)

lib/pundit.rb
```diff
@@ -8,7 +8,17 @@
 
 module Pundit
   class NotAuthorizedError < StandardError
-    attr_accessor :query, :record, :policy
+    attr_reader :query, :record, :policy
```
thomasklemm: Much cleaner interface for the `NotAuthorizedError`, looks great!

lib/pundit.rb
```diff
@@ -64,14 +74,19 @@ def verify_policy_scoped
 
   def authorize(record, query=nil)
     query ||= params[:action].to_s + "?"
+
+    if record.blank?
```
thomasklemm: This will cover a few more objects than `nil`. `'', ' ', nil, [], and {}` e.g. are all blank. I think that's OK here though, the error message below might be misleading however in case something other than `nil` has been passed in. Maybe we can clarify things a bit more in the error message, too, e.g. by saying something like "cannot authorize a blank object".

michaldarda: Agree - fixed

jnicklas:
Merged! Awesome, thank you!

davegudge:
+1 for https://github.com/elabs/pundit/issues/251#issuecomment-120315941

toshify:
Any chance you might reconsider this implementation? (See #251 for comments received after closing)

PR #254, opened 2015-02-26
Fix tests on activesupport v4.2.0

ecbypi:
`ActiveSupport` v4.2.0 deprecated `Kernel#capture` and uses
`ActiveSupport::Deprecation` to notify the user. Since spec_helper.rb
doesn't require "active_support" to setup autoloading, or
"active_support/deprecation" directly, this causes the call to `capture`
to fail. Since `Kernel#capture` is deprecated, copy (and loosely modify)
it's implementation.

Adding `require "active_support"` is still required to handle additional
uses of `ActiveSupport::Deprecation` in the rest of the library when
requiring `active_support/core_ext`.

Open to any feedback / desired changes.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,17 +1,8 @@
-require "active_support/core_ext/kernel"
-
-warnings = capture(:stderr) do
-  require "pundit"
-  require "pundit/rspec"
-end
-
-unless warnings.to_s.empty?
-  puts "ERROR: Encountered deprecation warning!"
-  puts warnings
-  exit 1
-end
+require "pundit"
+require "pundit/rspec"
 
 require "pry"
+require "active_support"
 require "active_support/core_ext"
 require "active_model/naming"
 


nitinstp23:
@jnicklas - Can this be merged?

jnicklas:
IIRC I later realized that there is a flag in RSpec which makes deprecation warnings raise an error. This would only affect RSpec obviously, but that's still more than we need. I'd rather see this entire mechanism removed than adding so much code for a silly safety net.

ecbypi:
@jnicklas OK. I'll update the pull request.

ecbypi:
@jnicklas, @thomasklemm I updated the pull request.

nitinstp23:
:+1:

radanskoric:
@jnicklas what is the "deprecation warnings raise an error" flag? I can't find it in the docs.

Can this be merged if that flag is added? This PR is needed so the other PRs can rebase and turn green.

radanskoric:
@jnicklas sorry to be a bother, but could you please have a look at this? It would be awesome to have this merged, a lot of other PRs need this to turn green. :)

Thanks for the hard work!

jnicklas:
I wan't able to find the flag in RSpec, I'm sure I've seen it somewhere, but it seems to be gone. Maybe it was only in some intermediate version. The whole 2->3 migration was a bit messy. Anyway, thanks for pinging me, I've merged this PR now.

radanskoric:
Awesome, thanks!

Issue #213, opened 2014-09-26
unable to find policy ArrayPolicy


acrogenesis:
Pundit is unable to find class for a CollectionProxy(Array) which contains several elements of a Model inside

darylprickett:
I think you might have a subtle security bug in your controller code. A single authorize check will determine if the controller action is valid. In your case you are calling authorize multiple times and I think your last result will determine if the action can run. Have you tested this using an array with mixed values like [true, true, false]?

acrogenesis:
@darylprickett Tried what you say and when a instance of authorize calls a raise it fails to update. To simplify the process I made a method on application_controller.rb

```ruby
def authorize_many(model_collection)
    model_collection.each do |model_instance|
      authorize model_instance
    end
  end
```

jnicklas:
`authorize` is meant to be used with a single object. I think that's well documented, and the fact that it doesn't work with an Array isn't really an "issue", so I'm closing this. Let's continue discussing this in #256 :)

PR #256, opened 2015-03-09
Authorize a ActiveRecord collection

acrogenesis:
authorize_many calls authorize on all members of the collection or array.

#213

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -76,6 +76,12 @@ module Pundit
 
     true
   end
+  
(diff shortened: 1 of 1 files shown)


jnicklas:
I think I'd be okay with this change, though maybe `authorize_all` or `authorize_each` would be nicer names? @thomasklemm, thoughts?

Anyway, this needs tests :)

jnicklas:
After considering this for a while longer, I'm no longer in favour of this. It adds more API surface for little gain.

PR #259, opened 2015-03-18
Add `pundit_policy_authorized?` and `pundit_policy_scoped?` methods

radanskoric:
In some cases we might need to simply check if `authorize` or
`policy_scope` were called, without immediately raising and exception.
In other cases we might be happy if either was called and only want
to raise if neither was called.

All those edge cases are most easily served by exposing lower level
function which just perform the checks and return `true` or `false`.

The `verify_*` methods can then build on top of them to offer a
convenient path for the most common use cases.

In my particular use case I am using pundit in a grape api application and the endpoints are not named like in a classical rails application so I can't simply use `verify_policy_scoped` on `index` actions and `verify_authorized` on other actions. What I really want to check is that either one was used because I the endpoints have very few lines of code and I just want to make it impossible to forget to use pundit policy altogether. Right now, I can't do that without relying on internal instance variables which could change with next pundit release. This PR fixes that.

### DIFF
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -224,6 +224,28 @@ describe Pundit do
     end
   end
 
+  describe "#pundit_policy_authorized?" do
+    it "is true when authorized" do
+      controller.authorize(post)
+      expect(controller.pundit_policy_authorized?).to be true
+    end
+
+    it "is false when not authorized" do
+      expect(controller.pundit_policy_authorized?).to be false
+    end
+  end
+
+  describe "#pundit_policy_scoped?" do
+    it "is true when policy_scope is used" do
+      controller.policy_scope(Post)
+      expect(controller.pundit_policy_scoped?).to be true
+    end
+
+    it "is false when policy scope is not used" do
+      expect(controller.pundit_policy_scoped?).to be false
+    end
+  end
+
   describe "#authorize" do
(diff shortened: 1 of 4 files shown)


jnicklas:
I'm okay with this change. It adds a bit of stuff which maybe isn't *that* generally useful, but nothing really complicated. @thomasklemm what do you think?

spec/pundit_spec.rb
```diff
@@ -214,6 +214,28 @@
...
+    end
+  end
+
+  describe "#pundit_policy_scoped?" do
+    it "is true when policy_scope is used" do
+      controller.policy_scope(Post)
+      expect(controller.pundit_policy_scoped?).to be true
+    end
+
+    it "raises an exception when policy_scope is not used" do
```
valscion: Exception? Probably you meant `is false` here.

radanskoric: Ups, :hankey: my bad, thanks for noticing it. I'll update it.

radanskoric: Updated and rebased! Thnx!

valscion:
If my input weighs at all, I'd say this PR is so small and easy to maintain in the future that I see no reason why this couldn't be merged.

jnicklas:
Okay, I'm merging this :)

radanskoric:
Thank you! :)

PR #260, opened 2015-03-27
Don't even look at model_name

jnicklas:
I was always under the impression that model_name's different variants (singular/plural etc...) are configurable, but this appears to not be the case. They simply delegate to the inflector. As such, there's really no point in having special cases for these. Tests still pass without this code.


twalpole:
the calls to model_name are needed for getting the class an ActiveRecord::Relation will return

jnicklas:
Ehh, of course! Never mind then :P

PR #261, opened 2015-03-27
Add permitted attributes helper, closes #141

jnicklas:
See discussion in #141. This provides a convenient helper which aids in permitting attributes in the controller.

PR #263, opened 2015-03-27
Add controller helpers which facilitate update and build patterns

jnicklas:
I had this idea that if #261 is merged, we would have a lot of code which looks like this:

```ruby
if @post.update_attributes(permitted_attributes(@post))
  redirect_to @post
else
  render :edit
end
```

Wouldn't it be nice if we could cut down on this noise?

```ruby
if update(@post)
  redirect_to @post
else
  render :edit
end
```

That looks much nicer, and makes our controllers a lot cleaner. We can also handle strange edge cases where it's unclear whether authorization should be performed before or after attributes are assigned by simply doing both.

I'm unsure if this is actually a good idea. It significantly expands how involved Pundit becomes in your application. Suddenly Pundit is not part of your controller actions, but rather actions are written in Pundit-style. This is one aspect I really *disliked* about CanCan, its `find_and_authorize` helpers just started to smell too much of `InheritedResources` or `ResourceController`, both of which I've been badly burned by. Is Pundit overstepping its bounds here, or is this inline with the kind of convenience we want to provide our users?

Also this PR does not (yet?) add separate `permitted_attributes` methods for `build` and `update`. IMO it should be possible to customize these individually. This is something we should do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -536,6 +536,49 @@ class PostsController < ApplicationController
 end
 ```
 
+## Controller helpers
+
+Pundit provides controller helpers which take the permitted attributes pattern to its
+logical conclusion. We can write an update action like this:
+
+```ruby
+# app/controllers/posts_controller.rb
+class PostsController < ApplicationController
+  def update
+    @post = Post.find(params[:id])
+    if update(@post)
+      redirect_to @post
+    else
+      render :edit
+    end
+  end
+end
+```
+
+Not only does this correctly retrieve the permitted attributes and assign them,
+it also makes sure authorization is performed both before *and* after we set
+these attributes. For example, an admin might be able to assign everyone a
+particular task, but a regular user can only assign themselves, not someone
+else. This kind of scenario requires the permissions to be checked twice, both
+before (they can't edit tasks assigned to someone else) and after (they can't
+assign a task to someone other than themselves).
+
+Similarly there is a helper for building a new record:
+
+```ruby
+# app/controllers/posts_controller.rb
+class PostsController < ApplicationController
+  def update
+    @post = build(Post)
+    if @post.save
+      redirect_to @post
+    else
+      render :edit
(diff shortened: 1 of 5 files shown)


valscion:
:-1: This is something I feel Pundit was really **not** designed to do and is overstepping its boundaries by a large amount. I like the current state of Pundit in that it allows **us** to write this kind of syntactic sugar in our **application code**, instead of a library with so much less visibility to our domain problems trying to tackle them for our benefit.

ecbypi:
@jnicklas `Pundit#update` needs a different name since the `update` action of the controller overrides `update` defined in `Pundit`.  Maybe `update_with_authorization`?

Your comments about cancan and inherited_resources makes me wonder if users will assume they should force their process through `update` and `build` instead of using them where they make sense which is what might have led to the bloated abstractions in those libraries.  If there is sufficient resistance to this then I think merging would be a good idea.

A nice middle-ground might be to include them as examples of abstractions in the README and defer to users to build them as needed in their own applications.  Or additional copy in the README indicating these helpers aren't intended to handle all use-cases and that in some cases, the user should use `authorize`, `permitted_attributes` and `policy` directly.  A good example of this is that `assign_attributes` will trigger database changes immediately when changing `has_and_belongs_to_many` and `has_many :through` associations on a persisted record.

jnicklas:
All feedback on this so far has been negative, which confirms my own unease about this feature.

The naming conflict that @ecbypi points out is also problematic, when I used this pattern in an application I had called the method `update_attributes`, but this doesn't have the same kind of symmetry with `build`, `build_attributes` does't have the same ring to it, and `update_with_authorization` as @ecbypi suggests doesn't sound nice to me.

I'm closing this PR for now. While I think it's an interesting idea, I think we need more experimentation and experience outside of Pundit for now.

Issue #212, opened 2014-09-23
Conversation about authorization, business logic, and custom error messages


uberllama:
Have been having an interesting conversation with some colleagues today and was wondering where you folks stand on it.

The question is: where do you draw your line between authorization and business logic/context? Do you use Pundit purely do define *who* can do *what* based on their role, or do you bring your deeper contexts into the picture?

Here's a plain english version of the scenario I'm dealing with:

Our system has a User model, a Post model, and an Event model. Users have roles ("admin" and "user").

* Post `belongs_to` Event, but can also stand alone.
* Any user of any role can create Posts not associated with an Event.
* Admin users can create Posts for any Event.
* Regular users can only create Posts for Events that haven't passed (if they try to create an Event for a past Post, they should get a custom error message).

### Option 1 Put everything in a Pundit Policy

If I were to do this all in Pundit, the PostPolicy and controller would look something like this:

```ruby
class PostPolicy

  def create?
    user.admin? || record.event_id.nil? || record.event.ends_at > Time.current
  end

end

class PostsController < ApplicationController

  before_action :set_post

  def create
    authorize @post
    if @post.save...
  end

  private

  def set_post
    @post = current_user.posts.build(post_params)
  end

end
```

This works... with caveats:

1. Pro: the policy could be used within a view to fully represent the rules around creating a Post.
2. Grey area: We're getting into tertiary data by looking at the record's Event association. This is a slippery slope.
3. Con: We don't have the ability to return a custom error message in the case of an expired Event (do we?). If authorization fails for whatever reason, you can only generate a singular error message based off the exception on `PostPolicy#create`.

### Option 2 Create multiple policies

```ruby
class PostPolicy

def create?
  record.event_id.nil? || record.event.ends_at > Time.current
end
```

```ruby
class AdminPostPolicy

def create?
  user.admin?
end
```

Assuming we could split policies cleanly along role lines, we might be able to do something like this, which opens up slightly more messaging options. Uncertain at this time, so this is the least fleshed out example.

### Option 3 Use Pundit purely for role checking (signed in, can create Posts) and put business logic in controllers

```ruby
class PostPolicy < ApplicationPolicy
  # ApplicationPolicy ensures user is logged in
end

class PostsController < ApplicationController

  before_action :set_post
  before_action :set_event

  def create
    authorize @post
    if @post.save...
  end

  private

  def set_post
    @post = current_user.posts.build(post_params)
  end

  def set_event
    return unless params[:post][:event_id].present?
[...]

PR #267, opened 2015-04-07
all Punit errors extend Punit::Error

TrangPham:
The reason behind this is because it would make it easier for users to rescue all errors from Punit using `rescue Punit::Error`

Please let me know if you think this is a bad idea. Personally, I have `rescue Pundit::NotAuthorizedError, Pundit::AuthorizationNotPerformedError` littered in my code whenever I want to rescue from StandardError but not any errors from Punit.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -9,7 +9,8 @@ require "active_support/dependencies/autoload"
 module Pundit
   SUFFIX = "Policy"
 
-  class NotAuthorizedError < StandardError
+  class Error < StandardError; end
+  class NotAuthorizedError < Error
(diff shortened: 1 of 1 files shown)


damien:
:+1: I've encountered similar problems trying to rescue errors specifically from Pundit. It actually complicated the implementation of #212 that one of my team members worked on. Having this in the mainline codebase would make extending Pundit and doing deep integrations with it safer/easier than it currently is.

jnicklas:
Sounds sensible to me! I usually set up error hierarchies this way, don't know why I didn't do it for Pundit.

PR #269, opened 2015-04-09
Add #skip_authorization helper_method

ltk:
The recently-added `#skip_authorization` method was not added as a helper, meaning the method is unavailable within Rails Controllers.

screenshot

There aren't any existing tests around these `helper_method` calls, so I followed convention and didn't try to rig anything up. Let me know though if you'd like me to add a tests to cover this.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -70,6 +70,7 @@ module Pundit
       helper_method :policy
       helper_method :pundit_policy_scope
       helper_method :pundit_user
+      helper_method :skip_authorization
     end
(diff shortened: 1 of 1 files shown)


jnicklas:
Yes, seems like we missed `skip_authorization`. But it should be marked with `hide_action` and not `helper_method`. `helper_method` makes controller methods available in views and `skip_authorization` is meant to be called from a controller, not from a view. The reason that `skip_authorization` didn't work for you is probably because it has not been released yet. The gem version doesn't have the method.

PR #277, opened 2015-04-23
Readme Update: NotAuthorizedError.initialize expects a hash

srecnig:
Please correct me if i did not understand something right, but I think i've found a mistake in your Readme.md: when initializing a NotAuthorizedError with just a string (like your Readme suggests) it will fail: 

    irb(main):001:0> raise Pundit::NotAuthorizedError, "must be logged in"
    TypeError: no implicit conversion of Symbol into Integer
	    from /Users/srecnig/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/pundit-1.0.0/lib/pundit.rb:17:in `[]'
	    from /Users/srecnig/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/pundit-1.0.0/lib/pundit.rb:17:in `initialize'
	from (irb):1

vs

    irb(main):010:0> raise Pundit::NotAuthorizedError, message: "must be logged in"
    Pundit::NotAuthorizedError: must be logged in

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ got through. This way you can fail more gracefully.
 ``` ruby
 class ApplicationPolicy
   def initialize(user, record)
-    raise Pundit::NotAuthorizedError, "must be logged in" unless user
+    raise Pundit::NotAuthorizedError, message: "must be logged in" unless user
     @user = user
     @record = record
   end


jnicklas:
At some point we changed the initializer of `Pundit::NotAuthorizedError`. I think we should allow passing a String to the initializer, so we should probably tweak the initializer to allow this form too.

srecnig:
since i just tripped over this again while upgrading on of our apps – can i help out with a for allowing to pass a string to the initializer?

jnicklas:
@srecnig yes, please!

jnicklas:
Closing this, since #279 has been merged.

PR #279, opened 2015-05-05
NotAuthorizedError can be initialized with a string

srecnig:
Like described in Readme.md, one should be able to initialize Pundit::NotAuthorizedError with a message string. See also: https://github.com/elabs/pundit/pull/277


jnicklas:
Nice!

srecnig:
:tada: glad i could help.

Issue #280, opened 2015-05-08
unable to find 'BetumPolicy' for ':beta'


jasonl99:
I must be missing something.  I modeled this after the DashboardPolicy example:

```ruby
class BetaPolicy < Struct.new(:user, :beta)
  def has_access?
    # determine access, return true or false
  end
end
```
When I do 
```ruby
authorize :beta, :has_access?
```

I get the following error:

```code
Pundit::NotDefinedError: unable to find policy `BetumPolicy` for `:beta`
from /home/jason/.rvm/gems/ruby-2.2.2/gems/pundit-1.0.0/lib/pundit/policy_finder.rb:30:in `policy!'
```

If I try calling
```ruby
authorize :betas, :has_access?
```
it works.   I imagine this has something to do with rails' thinking that beta is plural, and it's trying to singularize it, but I don't have a Beta or Betam table in the system.   Incidentally, using the DashboardPolicy example worked just fine.

deefour:
The Pundit policy finder uses `classify` under the hood. From the [Rails API docs](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-classify)

> Create a class name from a plural table name like Rails does for table names to models

Compare single/plural `:dashboard` to `:beta`

```
[7] pry(main)> :dashboard.to_s.classify
=> "Dashboard"
[8] pry(main)> :dashboards.to_s.classify
=> "Dashboard"
[9] pry(main)> :beta.to_s.classify
=> "Betum"
[10] pry(main)> :betas.to_s.classify
=> "Beta"
```

IMO, this should be considered a bug. When you're passing a symbol, it's to find a policy not bound to a table-based model.

jasonl99:
Thanks..I had just started looking into it myself and came to the same conclusion with classify.  Perhaps the better thing would be to use object.to_s.titleize.remove(' ')  I know there's a rails function in there somewhere which will do it with one method, though

jnicklas:
I think you are going about this the wrong way. The Rails inflector isn't pluralizing "beta" correctly, but instead of fixing *that* you're throwing out the baby with the bathwater.

``` ruby
"beta".classify # => "Betum"
ActiveSupport::Inflector.inflections(:en) do |inflect|
  inflect.irregular "beta", "betas"
end
"beta".classify # => "Beta"
```

I think Pundit is doing the right thing by using `classify`.

jnicklas:
[...]

PR #281, opened 2015-05-09
Fixes unexpected symbol-to-policy resolution described in #280

deefour:
When attempting to find a [headless policy](https://github.com/elabs/pundit#headless-policies), I think it would be better **not** to run the symbol through Rails' `classify` because the inflector singularizes the string.

Given some `CriteriaPolicy` and `PanelPolicy`, I think its better for the following behavior

```
Policy.find!(:criteria) #=> CriteriaPolicy
Policy.find!(:panel)    #=> PanelPolicy
Policy.find!(:panels)   #=> Pundit::NotDefinedError
```

instead of the current behavior

```
Policy.find!(:criteria) #=> Pundit::NotDefinedError (couldn't find a CriteriumPolicy class)
Policy.find!(:panel)    #=> PanelPolicy
Policy.find!(:panels)   #=> PanelPolicy
```

I'd prefer the symbol passed to Pundit to be cast to a string and camel-cased, but **not** singularized.

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -47,18 +47,14 @@ module Pundit
         elsif object.is_a?(Class)
           object
         elsif object.is_a?(Symbol)
-          classify(object)
+          object.to_s.camelize
         elsif object.is_a?(Array)
-          classify(object.join('/'))
+          object.join('/').camelize
         else
           object.class
         end
         "#{klass}#{SUFFIX}"
       end
-    end
-
-    def classify(name)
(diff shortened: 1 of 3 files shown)


deefour:
@jnicklas Thanks. I didn't think it through completely either and had just pulled `classify` from Rails minus the `singularize` bit. `camelize` on it's own is in fact simpler.

ryansch:
:+1:

jnicklas:
@deefour didn't see that you'd updated this. Merged. Thank you!

PR #282, opened 2015-05-14
Update to README.md

cblunt:
This change just updates the README with working code when calling `raise` or `fail` with a `Pundit::NotAuthorizedError` to supply the message as a hash

The old example code threw a `TypeConversionError` when using Pundit 1.0.0.


jnicklas:
Obsoleted by 1723a1b038d384c1d9a042c9c7c20958ebd349fd

PR #292, opened 2015-07-03
Try convert record #to_model in case we got decorated record

timurvafin:
In case we use decorated (presenter) object when passing to `policy` helper we should try convert in to model.


ce07c3:
Not too happy with introducing a view dep into an auth framework...`ActionView`.

jnicklas:
A definite no to this. Override `model_class` in your presenter if you really need this (probably a bad idea too), or just call `#to_model`.

PR #297, opened 2015-07-13
Add autodetection and override of param key for permitted_attributes helper

Envek:
First, trying to get param key for ActiveModel objects, as Rails' `form_for` helper uses it to generate key for parameters hash. See https://github.com/rails/rails/blob/9f44aa08636dfbba9261f4350ec14684425c4b7b/actionview/lib/action_view/helpers/form_helper.rb#L433 for reference.

Second, allowing to redefine it with `param_key` option (as element in optional second argument hash), it is useful in case of single table inheritance, when the param key is differs from actual class (as it is getting defined at runtime).

And last, use previous behaviour otherwise.

I believe that it fixes https://github.com/elabs/pundit/issues/286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Pundit
 
+- Add autodetection of param key to `permitted_attributes` helper with ability to override it with `param_key` option.
+
 ## 1.0.1 (2015-05-27)
 
 - Fixed a regression where NotAuthorizedError could not be ininitialized with a string.
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -141,10 +141,10 @@ module Pundit
 
   def permitted_attributes(record, options={})
     name = options.delete(:param_key)
-    name ||= if record.class.respond_to?(:model_name) # ActiveModel and ActiveRecord
-               record.class.model_name.param_key
-             elsif record.respond_to?(:model_name)
+    name ||= if record.respond_to?(:model_name) # ActiveModel and ActiveRecord
                record.model_name.param_key
+             elsif record.is_a?(Class)
+               record.to_s.demodulize.underscore
              else
                record.class.to_s.demodulize.underscore
              end
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -139,8 +139,15 @@ module Pundit
     policies[record] ||= Pundit.policy!(pundit_user, record)
   end
 
-  def permitted_attributes(record)
-    name = record.class.to_s.demodulize.underscore
+  def permitted_attributes(record, options={})
+    name = options.delete(:param_key)
+    name ||= if record.class.respond_to?(:model_name) # ActiveModel and ActiveRecord
+               record.class.model_name.param_key
+             elsif record.respond_to?(:model_name)
+               record.model_name.param_key
+             else
+               record.class.to_s.demodulize.underscore
+             end
     params.require(name).permit(policy(record).permitted_attributes)
   end
 
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -3,6 +3,7 @@ require "spec_helper"
 describe Pundit do
   let(:user) { double }
   let(:post) { Post.new(user) }
+  let(:customer_post) { Customer::Post.new(user) }
   let(:comment) { Comment.new }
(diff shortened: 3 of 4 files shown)


scottjacobsen:
I like this. I just noticed it after opening #299.

If you add one more leg to the if/else contraption I think we are on our way :)
```ruby
...
elsif record.is_a?(Symbol)
  record
else
  record.class.to_s.demodulize.underscore
...
```

jnicklas:
I like this, but I don't really see the point of sending in a param_key as an option. Can we get rid of that part?

Envek:
@scottjacobsen I think it's good idea to not only add Symbol support, but pick and adapt whole [class detection logic from policy finder's `if`] (https://github.com/elabs/pundit/blob/c5b3fc14050d6fee535804ff681d4d9cf0f2f9b9/lib/pundit/policy_finder.rb#L43-L55). Will update this PR in a few hours.

@jnicklas I have found overriding option useful when working with STI models, like that:

```ruby
class Attachment < ActiveRecord::Base; end
class Image < Attachment; end
```

And in our app when we creating attachment our params is always contains `attachment` key with hash with this attachment params, no matter of its type, so our controller action looks like this:

```ruby
attachment_type = Attachment.guess_type_for(params[:attachment][:attach][0]) # Some magic with MIME types is hidden there
@attachment = attachment_type.new
@attachment.attributes = permitted_attributes(@attachment, param_key: 'attachment')
```

It can be workarounded with simple `Attachment.new` and changing object type later though, but I think that possibility to override autodetected key may be useful under other circumstances as autodetection will not cover all the possible cases anyway.

Envek:
@scottjacobsen I've looked for it a bit further, and I don't think that it should be just “use this symbol as key” as it may require to find class for this symbol and then ask this class for its `param_key`.
Isn't it simpler to just pass a class into `permitted_attributes` helper?

But I can extract previously mentioned [policy finder if](https://github.com/elabs/pundit/blob/c5b3fc14050d6fee535804ff681d4d9cf0f2f9b9/lib/pundit/policy_finder.rb#L43-L55) to class level method to reuse it here.

scottjacobsen:
Okay. Best to leave the symbol part out of this one to avoid adding complication.

scottjacobsen:
I think I understand where you're going, but after thinking about this a bit I think it is a bit too much 'magic'. I tend to create a lot of decorators for my AR objects, and I've learned from experience that trying to manage the `model_name` can get out of hand. Also I don't think many people really understand `model_name`.

Also, the disconnect between the code the figures out the `name` value and the code that finds the policy seems like it could cause subtle bugs.

Would it make sense to make the strong params `require` key be an attribute on the policy? Something like this:

``` ruby
def permitted_attributes(record)
   policy = policy(record)
   name = policy.respond_to?(:permitted_attributes_key) && policy.permitted_attributes_key
   name ||= policy.class.to_s.demodulize.underscore.sub(/_policy^/, '')
   params.require(name).permit(*policy.permitted_attributes)
end
```

``` ruby
class SpecialAttachmentPolicy
  def permitted_attributes_key
    :attachment
  end
   
  def permitted_attributes
    ...
  end
end
```

As a bonus `permitted_attributes` could take a symbol :)

Envek:
@scottjacobsen, how do you propose to solve my example above when in controller's action we're creating `Image`, but in `params` we have `attachment` key (because we do not know type in advance). I think that only overriding option may solve this (or refactor all the controller action and client code) as, for example, in edit view `form_for` will generate `image` param key.

@jnicklas, do you have any thoughts?

scottjacobsen:
Thanks for the clarification I didn't understand the whole issue. 

If you were to set the name in the policy you would need to also override the parameter key in forms that use images. With strong params and form_for I wouldn't expect much other code would need to change:

```
<%= form_for(@image, as: :attachment) do |f| %>
```

``` ruby
class ImagePolicy
  def def permitted_attributes_key
    :attachment
  end
end
```

You can also override the method in your own application controller

``` ruby
class ApplicationController < ApplicationController::Base
  def permitted_attributes(record, name=nil)
    if name
      params.require(name).permit(*policy(record).permitted_attributes)
    else
      super(record)
    end
  end
end
```

jnicklas:
I've merged this without the option of overriding the param_key. It's better than nothing at least.

Issue #298, opened 2015-07-14
permitted_attributes always uses outer :require


stephendolan:
I'm using Pundit in my API, and was going to try to filter parameters based on user role.

My issue is that instead of my requests being like the following for updating a user:

```ruby
user: {
  first_name: "Test",
  last_name: "User"
}
```

They are formatted like this:

```ruby
{
  first_name: "Test",
  last_name: "User"
}
```

The default method for permitted_attributes in Pundit source is here:

```ruby
def permitted_attributes(record)
    name = record.class.to_s.demodulize.underscore
    params.require(name).permit(policy(record).permitted_attributes)
end
```

It would be great to have an options hash for this method that could specify something like not to use the `.require(name)` part.

openface:
+1 agree. 

Our API does not use root nodes either. We'll likely be hitting this same issue.

jnicklas:
Pundit is all about conventions, and having params nested under a key is, like it or not, a very strong convention in the Rails world. It's really easy to override `permitted_attributes` to do whatever you want in your controller. If you're building an API which uses root level params, that's what you should do IMO. We can bend over backwards trying to accomodate all possible ways people might use Pundit, but I don't think that's a good idea.

Rather see Pundit as a set of ideas, and if they don't suit you somehow, it's easy to override them.

PR #303, opened 2015-07-23
Add option for permitted_attributes to not require a root node.

stephendolan:
This pull request is in reference to #298.

It's my first open source contribution, so please let me know if I did anything dumb or not in line with the contribution guidelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,28 @@ class PostsController < ApplicationController
 end
 ```
 
+It is a Rails convention to require the outer object tag when permitting attributes, like so:
+
+```ruby
+params.require( :post ).permit( :title )
+```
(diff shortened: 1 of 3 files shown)


scottjacobsen:
See also #297 where adding an options hash to this method was proposed. `permitted_attributes` is surprisingly (to me) complicated. I'm against options hashes - methods should have few parameters and an options hash is a way to sneak lots of parameters into a method without looking like you are sneaking lots of parameters into a method.

stephendolan:
I definitely see where you're coming from. I suppose I can follow your advice in that thread and just override the `permitted_attributes()` myself in the application controller. Thanks!

PR #313, opened 2015-08-26
Use RSpec.describe in spec template

### LINKED ISSUE #311: RSpec generator should use `RSpec.describe`
The RSpec generator currently generates a spec which calls `describe` from the global object:

```RUBY
describe SomePolicy do
end
```

However RSpec 3+ disables the globals by default which gives the error `undefined method 'describe' for main:Object`.
Instead `RSpec.describe` should be used. This is also backwards compatible.

maxcal:
Fixes #311


john-griffin:
Also just hit this when generating a policy.

damien:
:+1:

jnicklas:
Thanks :smile:

PR #318, opened 2015-09-17
Added collection_authorizable method

olegantonyan:
In some cases I need to be able to authorize plain array of objects with some policy (for index action, for example). You have to define `policy_class` method on it. This PR adds helper method to do this.
```ruby
def index
  @collection = collection_authorizable [1,2,3], SomePolicy
  authorize @collection #will use SomePolicy class
end
```

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -61,6 +61,20 @@ module Pundit
     def policy!(user, record)
       PolicyFinder.new(record).policy!.new(user, record)
     end
+
+    def collection_authorizable(collection, klass)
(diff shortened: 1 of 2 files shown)


jnicklas:
This is a pretty broken way of doing this, IMO. You could just do:

``` ruby
@collection.each do |item|
  authorize(item)
end
```

Which we are not going to provide any sugar for.

olegantonyan:
But this will not work with empty collection, and you will get NotAuthorizedError in `after_action :verify_authorized`

PR #319, opened 2015-09-18
Update README.md

justin808:
1. Add example of passing the class, rather than an instance of the `record`.
2. Add example of passing 2 permissions for the test.


damien:
:+1:

jnicklas:
Thanks!

PR #330, opened 2015-11-20
ApplicationPolicy template: better defaults for edition/creation

vemv:
It's good to force policies to define e.g. either `new?` or `create?`, else bugs can happen.

### DIFF
--- a/lib/generators/pundit/install/templates/application_policy.rb
+++ b/lib/generators/pundit/install/templates/application_policy.rb
@@ -15,7 +15,7 @@ class ApplicationPolicy
   end
 
   def create?
-    false
+    new?
   end
 
   def new?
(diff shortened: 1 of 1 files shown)


deefour:
Pundit currently provides a sane, safe default, disabling access to creation/mutation.

Unless I'm missing something, you replaced this sane default with an endless loop.

vemv:
Choosing `create?` as the source for `new?` is quite arbitrary. Why is it not the other way around? Will the programmer remember that `new?` equals `create?` but not vice versa?

It's more cautious to let it fail (with a loop, yes) unless one overrides either `new?`or `create?` in his specific subclass.

Anyway, my PR is no big deal obviously :)

deefour:
I think you'll find the consensus is that pundit should function out of the box instead of throwing an app into an endless loop.

vemv:
Some people like fail fast. Many 'default' classes e.g. throw a NotImplementedError because they expect to be overriden.

No big deal!

Issue #308, opened 2015-08-21
Custom policy class resolver


pnomolos:
What are your thoughts on having the ability to specify a custom proc/lambda/class to resolve the classes for a given policy?  For instance, the current app I'm working on has two separate user classes - one for the front-facing website (`Agent`) and one for the administration section (`AdminUser`).  However, they do end up administering the same classes, which can make building policies tricky because I now have to check the class of a passed-in `user`.  Having a custom resolver would let me have completely different classes to handle each user-type, rather than resorting to a bunch of hacks, so I could have these:

- `class AdminUser::OrderPolicy ... end`
- `class Agent::OrderPolicy ... end`

As I'm writing this, I realize I'd be content if there was simply a setting to namespace or prepend the class name of the `user` to the policy lookup.  My initial solution is _probably_ a bit overkill (although still an interesting idea, I suppose - if you were using a role-based auth scheme and the roles got quite complex you could break down policies according to role).

Thoughts?  I'd be willing to write an initial implementation of either of these if the idea has merit.

jnicklas:
Namespaces have been suggested a bunch of times, but it's quite tricky to implement. We had a solution a while ago, but rolled it back because it didn't work properly in some edge cases. Someone also implemented a hacky fork of Pundit which does this, see https://github.com/coryodaniel/regulator, but I don't really understand how it's supposed to work.

As for a custom resolver, I think this is out of the scope for Pundit.

I'm not opposed to the idea of namespaces, in general, but the implementation needs to be airtight.

pnomolos:
@jnicklas regulator does seem like overkill for what I'm doing.  Right now I'm just hoping for: 
  - "Does one of the following exist?"
  - `Agent::QuotePolicy` or `AdminUser::QuotePolicy`
  - If not, then use `QuotePolicy`

Any existing namespace support would be tacked on, so (and only if this is already working) if you had `SomeModule::Agent` and `SomeModule::QuotePolicy` the resolution would be to `SomeModule::Agent::SomeModule::QuotePolicy`, falling back to `SomeModule::QuotePolicy`.  No attempting to figure out that `SomeModule` refers to the same namespace.

Alternatively, what are your thoughts on using DI for the `PolicyFinder` class, so a user can specify a custom policy finder on their own (without having to clobber the existing class, though)?

jnicklas:
[...]

PR #331, opened 2015-11-21
Better handling of policy finder for arrays and scoped models

ramaboo:
This pull request improves the ability of the policy finder to find policies from an array. This is particularly useful for a situation where you have a sub application within your main Rails application. A typical example of this is an Admin namespace.

```
module Admin
  class UserPolicy < ApplicationPolicy
    def show?
      true
    end
  end
end
```

With this pull request you can now do the following to locate this policy:
```
@user = User.new

policy([:admin, @user]).show?
policy([:admin, User]).show?
policy([:admin, :user]).show?
```

This saves having to namespace every model within an Admin module while still allowing you to have separate polices for the admin scope. You can of course still do `policy(Admin::User).show?` if you defined `Admin::User` class .

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -40,20 +40,26 @@ module Pundit
       elsif object.class.respond_to?(:policy_class)
         object.class.policy_class
       else
-        klass = if object.respond_to?(:model_name)
-          object.model_name
-        elsif object.class.respond_to?(:model_name)
-          object.class.model_name
-        elsif object.is_a?(Class)
-          object
-        elsif object.is_a?(Symbol)
-          object.to_s.camelize
-        elsif object.is_a?(Array)
-          object.join('/').camelize
+        klass = if object.is_a?(Array)
+          object.map { |x| find_class_name(x) }.join('::')
         else
-          object.class
+          find_class_name(object)
         end
         "#{klass}#{SUFFIX}"
+      end
+    end
+
+    def find_class_name(subject)
+      if subject.respond_to?(:model_name)
+        subject.model_name
+      elsif subject.class.respond_to?(:model_name)
+        subject.class.model_name
+      elsif subject.is_a?(Class)
+        subject
+      elsif subject.is_a?(Symbol)
+        subject.to_s.camelize
+      else
+        subject.class
       end
     end
   end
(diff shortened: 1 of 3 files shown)


guzart:
I think the right approach is to have the policy delegate to a scoped policy as in #308, this is only because Pundit is mostly a resourced based authorization. I believe all the controller should know, is that it needs to authorize the given resource for the current_user (`authorize @record`), adding knowledge of the policy scope feels out of place.

Great job on the PR, with tests and everything.

ramaboo:
@guzart In theory I agree with you. In practice it's not that simple. Working on a large codebase developed over many years things are often not in the exact place you would like them to be for guessing the scopes. Sometimes you want to treat `@user` as a normal user and other times as an Admin user or other times as an API user. Often times these decisions are made at the view layer as well which makes automatic scoping harder. You could of course add a property to determine which scope to check for the user object but then you would have to go that to every other object in the system. Is it a Tag or Admin Tag, etc, etc. In a large project that's just not practical.

If you look at the pull request I have only added functionality not removed it (and a little refactoring). Basically the only change is the find method used to guess the class name and if it got an array then it just used `[:symbol, :symbol]`. What I changed was looping though the array and guessing the class name on each item of the array rather than assuming they were all symbols. This lets you do lots of fun combinations of things to hint to pundit what scope you want.

jnicklas:
I've merged this PR. It's by far the best solution to the namespacing problem I've seen so far, and it's nice and simple and well tested. Thank you @ramaboo and apologies for not getting to this sooner.

ramaboo:
Thanks!

heliohm:
I find myself needing to DRY up my code when using namespacing with headless policies and arrays as arguments to **authorize**. As long as I keep my policies within the same folder structure as the controller namespace I can get away by including a helper module to build the needed array:

`module ModuleName

  class Account::DashboardController < BaseController
    include PolicyHelper

    def index
      authorize policy_namespace(:dashboard), :index?
    end
  end
end`

`module ModuleName
  module PolicyHelper

    def policy_namespace(object)
      @arg = self.class.to_s.split("::").map!{ |x| x.underscore.to_sym}
      @arg[-1] = object
      @arg

    end
  end
end`

I have not found a simpler way of doing this yet but I think it might be a bit more complicated than necessary..

PR #341, opened 2015-12-26
Better documentation [ci-skip]

asiniy:
* One line assignement
* We are all living in 2016 year, so ruby 2.4 is present. It's strange to look at 1.9.3 code style :)

Happy new year!

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ class PostPolicy
   attr_reader :user, :post
 
   def initialize(user, post)
-    @user, @post = user, post
+    @user = user
+    @post = post
   end
 
   def update?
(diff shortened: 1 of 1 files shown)


bolshakov:
It looks like commit message is not appropriate :)

jnicklas:
I don't particularly like one-line assignment. It's error prone and harder to read. Saving a line or two is not worth it, IMO. As for hash hash syntax, I'm okay with switching to JSON-style hashes (which btw were introduced in 1.9 ;)).

asiniy:
@jnicklas made it with small Tabularization!

jnicklas:
*sigh* I'm not happy you did this, and please don't do it again, but I can't be bothered to argue the point, so merged.

asiniy:
@jnicklas what's up? Why do you dislike tabularize?

jnicklas:
@asiniy that's not really the point. It's that changing the code style of someone else's project without asking them first is a bit... impolite.

asiniy:
Understood

PR #343, opened 2016-01-12
Add YARD docs for Pundit::PolicyFinder

bolshakov:
`Pundit::PolicyFinder` documented with YARD comments.

But I'm not sure if `PolicyFinder` and `PolicyFinder#find` is intended to be public interface. 
So I decided to clarify it using YARD tags (`@api public`).

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -1,17 +1,40 @@
 module Pundit
+  # Finds policy and scope classes for given object.
+  # @api public
+  # @example
+  #   user = User.find(params[:id])
+  #   finder = PolicyFinder.new(user)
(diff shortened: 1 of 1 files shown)


bolshakov:
I think it may be somehow related to #266

bolshakov:
I've just realized that this is not the right way to override policy class. To override policy class it's better to implement `AdminController#policy` method. 

But I hope this PR is still useful.

jnicklas:
Nice :D Thank you! Now that is the kind of PR that warms my heart :heart:

jnicklas:
screen shot 2016-01-14 at 13 39 51

Look, it's not entirely empty anymore!

bolshakov:
Yep, but there is a room for further improvements :)

PR #348, opened 2016-01-17
moving scopes higher in README

hwork:
This is a minor reordering of the README to introduce scopes before explaining how to ensure they (and policies) are used in the controllers.


jnicklas:
Awesome, thanks! This is indeed much better.

PR #350, opened 2016-01-18
Add options to policy scope

rafcamlet:
Now policy_scope method receive options. Dev can manualy set policy user
and/or namespace.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -83,8 +83,8 @@ module Pundit
   end
   attr_writer :policy_scope
 
-  def policy(record, options={})
-    namespace = options[:namespace] || self.class.parent
+  def policy(record)
+    namespace = self.class.deconstantize
     @_policy or Pundit.policy!(pundit_user, record, namespace)
   end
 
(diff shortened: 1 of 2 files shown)


jnicklas:
It seems like you've closed this PR from your end, nevertheless I wanted to leave some notes on how to send better PRs in the future:

- You should not touch the version number in PRs, the version number only changes when we do releases.
- I don't know what the change in `policy` is for, it seems unrelated. Always be sure to only include what you actually want changed in the PR.
- There are no tests. The contribution guidelines unequivocally state that PRs without tests will not be accepted. https://github.com/elabs/pundit/blob/master/CONTRIBUTING.md#pull-requests
- You did not add or update documentation for this feature as requested in the documentation guidelines.
- The PR seems to be based off an old version of the code. Always send PRs against the latest version, if the maintainers want to backport it, that is their choice.

Finally, before you invest time in sending a PR it is better to open an issue and ask the maintainers if they like or agree with this feature. In this case the answer would have been a clear "no" from my side. I do not think this feature is a good idea at all, so I wouldn't have accepted this PR anyway.

Sorry for the drawn out critique, but see this as things to improve next time rather than discouragement.

Most importantly: please read the contribution guidelines! They exist for a reason.

PR #352, opened 2016-01-20
use record#to_s in NotAuthorizedError#to_s and record#inspect in NotAuthorizedError#inspect

razum2um:
I read the https://github.com/elabs/pundit/pull/200 but I think, that `Exception#message` and `Exception#to_s` shouldn't include any sensible data

Imagine you're denying show? for some some record, but the error is getting rendered via #to_s somewhere => protected data will be exposed to anyone

Of course, rendering plain exception message to the end user is not the best idea,
but I think that default behaviour should protect anyone from such error.

This PR makes errors both:
- programmer-friendly as REPL uses #inspect when printing objects
- rendering-friendly, as record#to_s shouldn't include sensible data by design (as mentioned in #200)

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -28,11 +28,16 @@ module Pundit
         @query  = options[:query]
         @record = options[:record]
         @policy = options[:policy]
+        @inspect = "#<#{self.class}: not allowed to #{query} this #{record.inspect}>"
 
-        message = options.fetch(:message) { "not allowed to #{query} this #{record.inspect}" }
+        message = options.fetch(:message) { "not allowed to #{query} this #{record || "nil"}" }
       end
 
       super(message)
+    end
+
+    def inspect
+      @inspect || super
     end
   end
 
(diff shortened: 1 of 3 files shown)


jnicklas:
I don't think anyone really inspects errors, so this makes the error much harder to interpret.

I think providing a useful error is important, I think sending error messages to the user is fundamentally bad no one should do it.

I'm not sold on this.

jnicklas:
Also you probably mean "sensitive", rather than "sensible" ;)

damien:
I disagree on the topic of `inspect`—during development and debugging it's really nice to be able view an object without having to call methods on it to see what's going on.

That being said, I agree with @jnicklas that exceptions shouldn't hide data. It's generally not safe to assume you can show an exception to an end user—more often than not an exception will expose data about your system you don't want widely known. This is also why a lot of common exception tracking services provide mechanisms to filter out certain exception properties or metadata prior to sending them to an API.

thomasklemm:
I agree with @jnicklas and @damien that error messages should not be rendered directly from an exception. Instead you can (and probably should) [rescue a `Pundit::NotAuthorizedError`](https://github.com/elabs/pundit#rescuing-a-denied-authorization-in-rails) where you can e.g. set a flash message for the user, and maybe set [custom error messages via I18n](https://github.com/elabs/pundit#creating-custom-error-messages).

jnicklas:
I'm closing this PR. I'm sticking to my guns on this one. Readable error messages are important.

PR #354, opened 2016-01-28
Update README's Rescuing section

somazx:
I thought it worth mentioning this alternate approach to handling Pundit's NotAuthorized exception. With this approach the exception is handled by rails and the appropriate 403.html or dynamic generated error page is served.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,10 @@ class ApplicationController < ActionController::Base
 end
 ```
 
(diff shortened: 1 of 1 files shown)


jnicklas:
Neat! I didn't know this is possible.

somazx:
Yes, it took me a fair amount of googling to find that. Perhaps I should have included a reference to where I found it in the documentation: http://guides.rubyonrails.org/configuring.html#configuring-action-dispatch

PR #356, opened 2016-02-01
Add a little more context for verify_* methods

iamvery:
* These are sometimes misunderstood. (Maybe just me?) Hopefully this
  helps. :smile:

Is this legit?

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,24 @@ exception you can use the two lower level methods `pundit_policy_authorized?`
 and `pundit_policy_scoped?` which return `true` or `false` depending on whether
 `authorize` or `policy_scope` have been called, respectively.
 
+### Why is this important?
+
+Having a mechanism that ensures authorization happens allows developers to
+thoroughly test authorization scenarios as units on the policy objects
+themselves. The benefits include:
+
+* request tests are lean and focused on happy path behavior
+* unit tests execute much faster keeping your test suite speedy
+
+### What is this not?
+
+It's not a failsafe for authorization. Imagine a scenario where there exists a
+controller action that mutates your service, e.g. `PATCH /users/:id`.  Assuming
+the `update` action for the `UsersController` is not authorized, a request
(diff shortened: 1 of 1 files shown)


jnicklas:
Hmm, I agree that these methods seem to confuse people, but I'm not sure that this is a helpful explanation.

iamvery:
I spend some time trying to get the explanation right for awhile, but I'm also underwhelmed. Do you agree with it fundamentally? i.e. they're an aid to provide assurance so policies may be thoroughly tested in isolation. (Maybe you had other use cases in mind?)

iamvery:
Here is another shot at an explanation. Is this any better?

---
### Why is this important?

Having a mechanism that ensures authorization happens allows developers to
thoroughly test authorization scenarios as units on the policy objects
themselves. The benefits include:

* request tests are lean and focused on happy path behavior
* unit tests execute much faster keeping your test suite speedy

### What is this not?

It's not a failsafe for authorization. Imagine a scenario where there exists a
controller action that mutates your service, e.g. `PATCH /users/:id`.
Assuming the `update` action for the `UsersController` is not authorized, a
request would result in a 500 error due to `verify_authorized`. However the
update would already be persisted leaving a vector for updates being applied
without proper authorization.

damien:
@iamvery I think the description in your [last comment](https://github.com/elabs/pundit/pull/356#issuecomment-178265046) was pretty good. If @jnicklas doesn't feel it's useful in the README it may still be a good addition to the wiki or some other design document.

iamvery:
Thanks @damien. I just hope that it's read. I've found this distinction really important. It took me a long time to really appreciate the usefulness these methods.

jnicklas:
@iamvery especially the last paragraph is great! That's definitely a very good explanation. Let's add this!

iamvery:
great! updated :smile_cat:

jnicklas:
Thanks!

PR #357, opened 2016-02-02
Replace calls to `hide_action` with protected access modifier

iamvery:
Curiosity got the best of me. After [striking up a conversation](https://github.com/elabs/pundit/pull/23#issuecomment-178278622) about `hide_action`, I decided to read up a little more about it in Rails. As it turns out, this feature [has been removed](https://github.com/rails/rails/pull/18371) from Rails master (Rails 5 beta). In order to maintain compatibility with future versions of Rails, I'm proposing that the use of `hide_action` be replaced with access modifiers in Pundit.

A couple things:

* Using `protected` over `private` so that subclasses of the including class may still access these methods.
* Since these methods are called directly in test, we raise the access modifier to public for the test object to preserve the test behavior.

### Concerns

1. Technically this changes the public API of the library. Arguably, it was never intended for these methods to be used outside of the object's including Pundit, but it's possible. Therefore, this is a breaking change (but maybe a "safeish" one :innocent:).
2. The tests that include a `Controller` feel somewhat misleading. The truth is, this is the public API of Pundit, but it should be internal to the including object...

I wanted to avoid large changes to the test suite and code for this proposal. Perhaps with some discussion, we can find a satisfactory solution :smiley_cat:

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -128,22 +128,9 @@ module Pundit
       helper_method :pundit_policy_scope
       helper_method :pundit_user
     end
-    if respond_to?(:hide_action)
-      hide_action :policy
-      hide_action :policy_scope
-      hide_action :policies
-      hide_action :policy_scopes
-      hide_action :authorize
-      hide_action :verify_authorized
-      hide_action :verify_policy_scoped
-      hide_action :permitted_attributes
-      hide_action :pundit_user
-      hide_action :skip_authorization
-      hide_action :skip_policy_scope
-      hide_action :pundit_policy_authorized?
-      hide_action :pundit_policy_scoped?
-    end
   end
+
+protected
 
   # @return [Boolean] whether authorization has been performed, i.e. whether
   #                   one {#authorize} or {#skip_authorization} has been called
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -153,6 +153,8 @@ end
 
 class Controller
   include Pundit
+  # Mark protected methods public so they may be called in test
+  public(*Pundit.protected_instance_methods)
 
   attr_reader :current_user, :params
 


damien:
This looks pretty reasonable to me. I like that the exposition of protected methods on the test object is as straightforward as it is. :)

iamvery:
Thanks @damien. I'm just a little put off by this being [action at a distance](https://en.wikipedia.org/wiki/Action_at_a_distance). At least that's isolated to the test? It's also weird that these things are the "public" API of Pundit (the pieces that users are expected to use), so maybe that feels weird to have them non-public in the lib? Hmm...

damien:
@iamvery I don't disagree with you, but I suspect it's going to be hard to move away from that when the expected usage is to include the root namespace of the library into a controller.

If we really want to enforce a strict API spec to be used within a controller, it may be worth putting helper-specific methods into their own module (ala `Pundit::Helpers` or somesuch) and update documentation to direct people to include that instead. Not exactly backwards compatible, but it'd be one way of avoiding inheritance of methods outside of the contexts for which they were intended.

iamvery:
Yeah it seems like it may be possible to split the large `Pundit` module up into smaller, singularly responsible pieces. This feels more like a design decision I should leave up to the project owner. Happy to help if that's desirable!

jnicklas:
I'm merging this as it feels very reasonable to me. As far as breaking changes go, this one feels pretty safe, nevertheless we should probably bump the major version for the next release. There is another minor breaking change I'd like to make in that case.

We could definitely factor the Pundit module into a lot of small helpers and test those individually, but that doesn't really feel nice. Changing the tests to not access pundit helpers from outside also seems to be pretty cumbersome, since that would mean a lot of indirection in the tests. We could use `send` to work around the visibility issue if we don't like the "hack" implemented in this PR, but imo it's not so bad.

iamvery:
Thanks for the reply @jnicklas. Glad you found this helpful :heart:

Onumis:
@jnicklas This is something I'm looking for, is there date for a new release? I see there's lot of new stuff since v1.1.0

PR #364, opened 2016-02-25
Fix param names in documentation

rob-murray:
Hi,
Great library, I've just started using it in a project with great results.

As I was reading the documentation I saw a couple of param names in the docs did not match those in the code so this just updates the documentation in a few places.

Hopefully any wording changes I've made describe the intention OK.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -56,7 +56,7 @@ module Pundit
     #
     # @param user [Object] the user that initiated the action
     # @param record [Object] the object we're checking permissions of
-    # @param record [Symbol] the query method to check on the policy (e.g. `:show?`)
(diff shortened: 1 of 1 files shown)


jnicklas:
Hmm, merged this, but just realized that one of the changes you made is incorrect. `it` refers to `permitted_attributes_for_{action}`, and not `action`. So defining, e.g. `permitted_attributes_for_update` would make that method be called rather than `permitted_attributes`.

rob-murray:
Yes, good point - should be something like "if the method is defined".

rob-murray:
Would you like me to create a new PR to change that line? I could also do any other docs changes you had in mind.

PR #383, opened 2016-04-22
Add splat lost in merge

MatthewRDodds:
In order versions of rails it wasn't possible to pass an array to `permit`, which is what `permitted_attributes` [returns](https://github.com/elabs/pundit/blob/master/lib/pundit.rb#L246).

In reference to: https://github.com/elabs/pundit/issues/353

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -243,7 +243,7 @@ protected
     else
       "permitted_attributes"
     end
-    params.require(param_key).permit(policy.public_send(method_name))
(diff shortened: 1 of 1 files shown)


jnicklas:
Thanks :)

adie:
Can you please release a new version with that fix? Because latest v1.1.0 is still affected by that problem: https://github.com/elabs/pundit/blob/v1.1.0/lib/pundit.rb#L255

Issue #375, opened 2016-03-14
Make authorize! return the object that was passed into the method.


jnicklas:
Right now it always returns `true`, this is somewhat useless. By having it return the object, we can do all kinds of useful patterns, like `render json: authorize(@foo)`.

ce07c3:
I would suggest that the policies themselves can return more than just true. One use case is evaluating rules in the policies - sometimes you need to know the results.

formigarafa:
@ce07c3 , I don't think that would be a good idea. We could end up with another service library doing lots of stuff instead of what it is supposed to do: check policies. That's exactly the reason of policy methods end in question marks. But if you think you need something more, take a look at active_interaction , that might give you what you are asking for here. :)

ce07c3:
@formigarafa There lies a lot of value in returning more than just true or false once your application needs grow. For instance, what if I wish to whitelist parameters for responses to API requests? If my policy would return a garafa result including the set of permitted attributes, I could pass this on to subsequent serializers. :)

formigarafa:
@ce07c3,
but that is exactly what a meant by doing too much stuff. It would become unclear what pundit is supposed to do. And while it might seem a good idea on a initial moment, when your application grow it will be even more difficult reuse a police/permitted-params thing. And like I said, you can still use a third class to achieve what you suggested. 
A PolicyAwareParameters class could get a policy instance and return your whitelisted params. But I think it makes more sense a class having pundit as a dependency to do something more than change it to have more responsibilities than any class should. This would probably lead to a future extraction of policy permission class from the policy class to ensure each one have a single responsibility.

BTW, I like your idea of having something to return the permitted params according to a policy.
It would be very helpful.

ce07c3:
I saw code or issues about caching policies. Since policies can only return true or false, I'd have to set data on the policy itself if I were to create a `PolicyAwareParameters` class, rendering them thread unsafe.

Also, I'd like to adhere to DRY principles so specifying the attributes in the policy feels right.

*active_interaction* looks nice, I came up with a few ideas on where to use it in my app (not for authorization, though =D)

fagiani:
@jnicklas when is this planned to get released?

PR #385, opened 2016-04-25
Update #authorize to return the passed object

aapomm:
Addresses the enhancement #375

- Replace the `authorize` return values with `record`
- Updated README.md with the change

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,15 @@ def admin_list
 end
 ```
 
+`authorize` returns the object passed to it, so you can chain it like this:
+
+Controller:
+```ruby
+def show
+  @user = authorize User.find(params[:id])
+end
+```
+
 You can easily get a hold of an instance of the policy through the `policy`
 method in both the view and controller. This is especially useful for
 conditionally showing links or buttons in the view:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -151,6 +151,16 @@ def admin_list
 end
 ```
 
+It is important to note that `authorize` returns the object passed to it,
```
jnicklas: I disagree that it is "important". I'd maybe say "kinda neat", but not "important" ;).

jnicklas: How about something like:

> `authorize` returns the object passed to it, so you can chain it like this:

aapomm:
@jnicklas I updated the wording in the README. Thanks!

martin308:
Is there a new release planned for this? Right now following the readme doesn't work at all due to this change

md5:
I hope nobody minds me bumping this thread. I'd also love to see this in a release. Looks like there is a "2.0" milestone planned.

@jnicklas Is there any chance of this making it out as a 1.x release or should we expect to wait until 2.0 is ready?

cakejelly:
Also would like to see this in a release but doesn't seem likely, based on the comments in https://github.com/elabs/pundit/issues/276

Seems strange though to merge into master and have it sitting there if there are no plans for a release 🤔

Issue #381, opened 2016-04-16
PolicyFinder to returned different Policies for Instance vs Class records


byu:
Feature request.

I'm noticing that I'm doing quite a bit of instance type checking when working with custom permission checks methods. Eg `record.is_a?(Post)` before calling methods on the record for the actual permission checks.

It would be easier for me to do the check just once in the initializer, eg,

```
def initialize(user, record)
  raise Pundit::NotAuthorizedError, 'Invalid Record Type' unless record.is_a?(Post)
  super
end
```

But that would cause problems for class scoped permission checks such as `create?`.

I'm thinking that having the ability to separate the concerns into two policy classes would make my code more assertive, mentally cleaner?

jnicklas:
Hmm, assertive code is good, I agree. I haven't really thought too much about how Pundit can be made more assertive. I suppose one way is to use the inference backwards, e.g.

``` ruby
def initialize(user, record)
  raise Pundit::NotAuthorizedError, 'Invalid Record Type' unless is_a?(Pundit.policy(record))
  super
end
```

But this is a little silly, imo.

scottjacobsen:
Generally type checking with `is_a?` is a bad code smell.

If you really need a different policy for `authorize(@post)` vs `authorize(Post)` I believe you could do something like this:

```ruby
class Post
  def self.policy_class
    PostClassPolicy
  end

  def policy_class
    PostPolicy
  end
end
```

scottjacobsen:
Re #387

PR #387, opened 2016-04-30
Add documentation for defining policy_class instance method

aaackerman:
I encountered a use case which required me to to define `policy_class` for instances of a class and had to do a little digging to make sure that I was able to do so. I thought it would be useful for other users of Pundit to see this feature via the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,16 @@ class Post
 end
 ```
 
+You can also explicitly declare which policy to use for a specific instance of a class.
+
+```ruby
+class Post
+  def policy_class
+    published? ? PublishedPostPolicy : PostPolicy
+  end
+end
+```
+
 ## Just plain old Ruby
 
 As you can see, Pundit doesn't do anything you couldn't have easily done


README.md
```diff
@@ -355,6 +355,16 @@ class Post
 end
 ```
 
+You can also explicitly declare which policy to use for a specific instance of a class.
+
+```ruby
+class Post
+  def policy_class
```
scottjacobsen: I'm fairly sure the proper way to do this is to define `policy_class` as a class method like so:

```ruby
class Post
  def self.policy_class
    ...
```

The reason is you can pass the class to the pundit helper methods - `authorize(Post)` for example.

aaackerman: Yes–that's definitely the main use case for `policy_class` @scottjacobsen, as described in the section above my addition. Defining the instance method seems more for edge cases where you want an instance to use a specific policy based on some attributes, say in a view helper `policy(post).publish?`, for example.

scottjacobsen: Oh, you're right. I wasn't reading carefully.

jnicklas:
I ... didn't know that this works. I'm not 100% sure we actually want to advertise this as a feature, since the fact that it works is a bit coincidental.

aaackerman:
Fair enough @jnicklas, I discovered reading [this line](https://github.com/elabs/pundit/blob/381a39e4923838257e247d4765ca39f9c10c8f7d/lib/pundit/policy_finder.rb#L79). Shall I go ahead and close the PR?

macfanatic:
This proves to be pretty helpful when you have complex permissions logic and turn to inheritance as a structure to make comprehension even easier.

I believe this always works, as the code is simple:

```ruby
# policy_finder
def find
  if object.respond_to?(:policy_class) # instance method
   object.policy_class
  elsif object.class.respond_to?(:policy_class) # class method
    object.class.policy_class
  end
  # more logic...
end
```

I had to browse through the source code and also saw #381 referencing exactly what we're wanting to accomplish.

Issue #351, opened 2016-01-18
Issue with new namespace finding capabilities (using an array)


zach-taylor:
When calling various helper methods in Pundit e.g. `policy([:admin, @user]).show?`, the full array is being passed to the policy object, rather than just `@user`. Is this expected functionality? or should my policy objects be grabbing the last object in the array to use as the `record` in the policy.

zach-taylor:
Better example:

```ruby
# app/controllers/admin/users_controller.rb
module Admin
  class UsersController < ApplicationController
    def show
      @user = User.find(params[:id])
      authorize([:admin, @user])
    end
  end
end

...

# app/policies/admin/user_policy.rb
module Admin
  class UserPolicy
    attr_reader :user, :record
    def initialize(user, record)
      @user = user
      @record = record # => [:admin, @user]
    end
  end
end

```

So what I'm having to do in my policy is this:

```ruby
@record = record.is_a?(Array) ? record[-1] : record

**Comment by jnicklas** [comment]
Hmm, that's not optimal. It seems somewhat unexpected to send through only the last item as well though. Damn.

**Comment by zach-taylor** [comment]
Yeah, not sure what the right solution is. @ramaboo any ideas?

**Comment by ramaboo** [comment]
@zach-taylor 

At my company we solve this problem in the `ApplicationPolicy` like this (which all other policies inherit from):

```
class ApplicationPolicy
  attr_reader :user, :record

  def initialize(user, record)
    @user = user
    @record = record.is_a?(Array) ? record.last : record
  end
end
```

I think the bigger question is should this be done at the Pundit level where record is always a single object or should there simply be better documentation around the need to do this.

**Comment by Burgestrand** [comment]
This also affects the test helpers, since you'll now want to pass in the full array to reproduce the actual conditions: `expect(policy_class).to permit(user, [:namespace, :here, record])`

**Comment by ramaboo** [comment]
What is the thinking on this. Should record aways just return the record or an array? I am working on some more pundit improvements so figured I could get this one in as well if people think it should.

**Comment by jnicklas** [comment]
So I've thought about this a bit more. I feel pretty strongly that we should pass only the *last* object into the policy. The reason for this design choice is that there is already precedent for exactly this kind of behaviour in Rails' forms. Consider this:

``` erb
<%= form_for [@post, @comment] do |f| %>
  <%= f.text_field :body %>
<% end %>
```

[...]

PR #391, opened 2016-05-05
Fixes namespace handling for policy_scope

ramaboo:
As discussed here https://github.com/elabs/pundit/issues/351 this fixes the policy_scope finder to work with namespaces.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -88,7 +88,8 @@ module Pundit
     # @raise [NotDefinedError] if the policy scope cannot be found
     # @return [Scope{#resolve}] instance of scope class which can resolve to a scope
     def policy_scope!(user, scope)
-      PolicyFinder.new(scope).scope!.new(user, scope).resolve
+      model = scope.is_a?(Array) ? scope.last : scope
+      PolicyFinder.new(scope).scope!.new(user, model).resolve
     end
 
     # Retrieves the policy for the given record.
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -87,7 +87,7 @@ describe Pundit do
       end.to raise_error(Pundit::NotDefinedError, "unable to find policy scope of nil")
     end
 
-    it "returns an instantiated policy scope given an array of a symbol and plain model class", focus: true do
+    it "returns an instantiated policy scope given an array of a symbol and plain model class" do
       expect(Pundit.policy_scope!(user, [:project, Post])).to eq :read
     end
 
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -86,6 +86,14 @@ describe Pundit do
         Pundit.policy_scope!(user, nil)
       end.to raise_error(Pundit::NotDefinedError, "unable to find policy scope of nil")
     end
+
+    it "returns an instantiated policy scope given an array of a symbol and plain model class", focus: true do
+      expect(Pundit.policy_scope!(user, [:project, Post])).to eq :read
+    end
+
+    it "returns an instantiated policy scope given an array of a symbol and active model class" do
+      expect(Pundit.policy_scope!(user, [:project, Comment])).to eq Comment
+    end
   end
 
   describe ".policy" do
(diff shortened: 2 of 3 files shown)


lib/pundit.rb
```diff
@@ -88,7 +88,8 @@ def policy_scope(user, scope)
     # @raise [NotDefinedError] if the policy scope cannot be found
     # @return [Scope{#resolve}] instance of scope class which can resolve to a scope
     def policy_scope!(user, scope)
-      PolicyFinder.new(scope).scope!.new(user, scope).resolve
+      model = scope.is_a?(Array) ? scope.last : scope
```
jimryan: Note that this will break if the scope is a Rails 3 relation.  See #402

ramaboo: Interesting... It's been so long since I used Rails 3 I had not considered this.

jimryan: It's very unfortunate behavior.  I think the only option might be to check `#model_name`, or drop support for Rails 3 (not my personal favorite choice, since I still maintain a Rails 3 application, but it would be understandable since even Rails doesn't support Rails 3 anymore).

Edwardzyc:
What's the status on this?

ramaboo:
Not sure we have been using this in production at Doorman.co for months now without issue. I would vote to drop rails 3 support as rails 5 has been released and supporting 3 versions is just crazy.

ce07c3:
@ramaboo yes, please drop Rails 3

chevinbrown:
+1

brunowego:
+1

stephen-puiszis:
👍

chevinbrown:
@ramaboo Can close & merge this PR with a readme update that explains the version w/ rails 3 support and that future versions will drop R3 support?

medir:
+1

medir:
After reading the discussion #351 I found very interesting the @mysmallidea implementation and give the idea to apply the same fix to the scope class.

```ruby
class ApplicationPolicy
  #index? show?....

  class Scope
    attr_reader :user, :scope

    def initialize(user, scope)
      @user = user
      @scope = scope.is_a?(Array) ? scope.last : scope #This fix the problem
    end

    def resolve
      scope
    end
  end
end

class Backoffice::EventPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      if user.is_admin?
        scope.all
      else
        scope.not_published
      end
    end
  end
end

class Backoffice::EventsController < ApplicationController
  def index
    authorize [:backoffice, Event]
    @events = policy_scope([:backoffice, Event]).page params[:page]
  end
end
```

chevinbrown:
@medir, we've had to do something similar in order to make namespaces work. Had a `#TODO` in that module for quite a while...Can confirm your solution is good for all our use cases.

rafaelsales:
Man.. this is really getting in my way - can we bring this thread back to life?

Linuus:
@ramaboo A long overdue reply here :) So, this fixes the issue that the whole array is passed to the policy scope initializer by passing only the last item, right? So we need to make the same fix for the Policy class as well, as discussed in #351 

I don't mean that you should add that fix in this PR, I just want to make clear that I understand what's going on here. I will wait with the merge here until I have a PR to fix the same issue for policies.

As this is a breaking change it will have to go into 2.0.

laertispappas:
Is anyone going to merge this fix?

laertispappas:
@Linuus Ack sorry

Linuus:
@ramaboo Could you rebase this on master and fix the conflicts?

lib/pundit.rb
```diff
@@ -88,7 +88,8 @@ def policy_scope(user, scope)
     # @raise [NotDefinedError] if the policy scope cannot be found
     # @return [Scope{#resolve}] instance of scope class which can resolve to a scope
     def policy_scope!(user, scope)
```
Linuus: The `policy_scope` method on line 78 needs to be updated as well.

Linuus:
Since I've not heard anything here, I added a PR which incorporates these changes and fixes the same issue for policy.

See #529

PR #393, opened 2016-05-10
clean application_policy template

ShPakvel:
Hi guys!

Here some thoughts I want to share with you.

1. The `scope` method is not used anywhere, but only here. It is not described in readme either. Which is strange. I would say if we have it in default installed base class then it should be something commonly used and should be at least mentioned as something useful in readme. I don't see any real usefulness of it comparing to other possibilities to work with scope. Also it is not common for all standard actions - for index we probably will pass class as record, so `record.class` in this `scope` method will give as not what we expect. Ok, then if it is not so common and so really especially useful outside of this class and not even mentioned anywhere then I can conclude that this is more for internal usage, mainly for `show?` method and should be private. But it is not private as well... All of this lead me to thought if it is really necessary here at all, which put our attention to `show?` method which is only who use it.

2. About `show?` method. Again, what I expect from default installed base class to do most possible thing which will work by default for all children. And yes, it is looks like so for me. But, what does it do actually? It takes record id and do request to DB to check existence of this record... But doesn't in general (let say in default way) taking id of the passed record tells us that we already have the record from DB? I believe so. Then I don't understand of necessity of this double check in `show?` method. More over it is additional (useless) query to DB which is not good for performance. So better we do check like `record.try(:id).nil?` here if you wish. But...

3. much better to return `false` for `show?` method by default as for other methods. I would say so cause it is always more secure to work with white list approach and disallowed status from the beginning. And I believe this library specialized on security, so I would say it is must be like this for it.

Hope it will make sense for you as for me.
And thank you for great gem :bow:

### DIFF
--- a/lib/generators/pundit/install/templates/application_policy.rb
+++ b/lib/generators/pundit/install/templates/application_policy.rb
@@ -11,7 +11,7 @@ class ApplicationPolicy
   end
 
   def show?
-    scope.where(id: record.id).exists?
+    false
   end
 
   def create?
@@ -32,10 +32,6 @@ class ApplicationPolicy
 
   def destroy?
     false
-  end
-
-  def scope
-    Pundit.policy_scope!(user, record.class)
   end
 
   class Scope


jnicklas:
I actually agree. I always end up removing this from applications I build. Maybe we should mention this in the README as a neat trick, but it shouldn't be the suggested default.

jzinn:
Even though it's been removed from the template, and I may also remove it from my application code, this is what I have currently:
```ruby
def scope
  self.class::Scope.new(user, object.class).resolve
end
```

My current implementation is because I had a problem with `Scope` class lookup due to policies being nested in module namespaces.

While investigating, I noticed that the code does the same policy finding twice, once from `authorize` in the controller and once from `policy_scope!` in the `scope` method.  Inside the `scope` method, `self` is already the policy object, so I used that instead of calling `policy_scope!`.

The `object` reference is set earlier in `initialize` like this:
```ruby
object = record.is_a?(Array) ? record.last : record
```
following how Rails' `FormHelper` [does it][1].

[1]: https://github.com/rails/rails/blob/v5.0.0.1/actionview/lib/action_view/helpers/form_helper.rb#L437

PR #394, opened 2016-05-20
don't let `include Pundit` override existing definition of `pundit_user`

asross:
I ran into this issue when trying to create a wrapper module to Pundit that I included in certain of my controllers where I didn't have access to a `current_user`, and there was a different session-specific object I needed to pass to Pundit instead. The docs suggested I should define `pundit_user` for this purpose, so I did, but it was never called because including Pundit actually overrode my definition of it (based on the module `include` ordering).

This updates the default definition of `pundit_user` to call the existing method if it's defined.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -266,6 +266,7 @@ protected
   # @see https://github.com/elabs/pundit#customize-pundit-user
   # @return [Object] the user object to be used with pundit
   def pundit_user
+    return super if defined?(super)
     current_user
   end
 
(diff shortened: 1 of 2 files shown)


jnicklas:
Why would you include `Pundit` in the `included` block instead of simply including it into the module?

jnicklas:
(which would put Pundit higher up in the inheritance chain, thus solving this problem)

jnicklas:
I.e. change your code to this:

``` ruby
      wrapper_module = Module.new do
        extend ActiveSupport::Concern
        include Pundit

        def pundit_user
          "Thomas Friedman"
        end
      end
```

asross:
@jnicklas ...that works 😅  Looks like I misunderstood modules.

This pull request still might save people a little bit of trouble if they make the same mistake, and it's probably harmless, but feel free to close or merge as you prefer.

Issue #340, opened 2015-12-07
Different permitted parameters per action?


RST-J:
Hello,
I'm working with pundit for the first time and like the [`permitted_attributes helper`](https://github.com/elabs/pundit#strong-parameters) as it bundles together (write) access control for actions and attributes.
However, I now encounter the situation where I want to allow different sets of parameters for different actions.
Would you consider adding such a feature if I prepared a pull request?

My idea is to extend permitted_attributes this way:
```ruby
def permitted_attributes(record)
  name = record.class.to_s.demodulize.underscore
  action_params_method = "#{params[:action]}_attributes"
  attributes_method = policy(record).respond_to?(action_params_method) ?
    action_params_method : :permitted_attributes
  params.require(name).permit(*policy(record).send(attributes_method))
end
```

Silex:
#299 should be considered if you refactor this.

jnicklas:
I'm reading this as that the method would have the name `create_permitted_attributes` for example? That seems a bit misleading to me. How about something more like `permitted_attributes_for_create`, I think that makes it clearer. Principally I think this is a neat addition :+1:

PR #395, opened 2016-05-24
Document permitted_attributes_for_#{action}

md5:
It looks like the changes made for #340 were not documented, so here's a proposed update for the `README.md`.


jnicklas:
Thanks :)

PR #397, opened 2016-06-07
Remove 10 extra spaces from the middle of RSpec failure messages

jzinn:


jzinn:
Rubocop passed when I ran it locally.  However, I tried and failed to get Rubocop to complain by creating a too-long line.

PR #401, opened 2016-06-24
Allow specification of a NilClassPolicy

bendilley:
This saves having to check whether objects are nil before checking their
policy.

This contradicts the intent of elabs/pundit#251 and could represent a
breaking change for some.

@jnicklas I hope you'll consider my reasoning here - I am using this branch in a project for checking permissions on associations which can legitimately be nil.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,27 @@ class ApplicationPolicy
 end
 ```
 
+## NilClassPolicy
+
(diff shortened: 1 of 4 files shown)


Linuus:
@bendilley Hi! Thank you for the contribution. Can you rebase this on master before I review it please?

PR #402, opened 2016-06-29
Fix policy lookup for Rails 3 relations

jimryan:
Rails 3 relations look like Arrays (see: https://github.com/rails/rails/blob/v3.2.22.2/activerecord/lib/active_record/associations/collection_proxy.rb#L25), but can be identified by checking if they respond to #model_name.

Originally, when a policy class was being looked up, the object's `model_name` was tried first, then the lookup code checked if the object was an `Array`.  This order was swapped in 4f22dc1, which broke compatibility with Rails 3 relations by first checking if the object was an `Array`, which caused Rails 3 relations to be treated as an array of model instances.  For example, if a Foo has 3 Bars: `policy_scope(@foo.bars)` would try finding the `Bar::Bar::BarPolicy` instead of the `BarPolicy` as intended.

This commit restores the order of those two checks.  One small side-effect of this commit: Array lookups are now recursive, ex: `Pundit.policy(user, [[:foo, :bar], Post])` is now equivalent to `Pundit.policy(user, [:foo, :bar, Post])` and both will instantiate the FooBarPost policy.  I didn't write an explicit test for this as I can't think of a reasonable use-case for this behavior, but also don't believe it should break existing code as I can't imagine someone would prefer that the nested Array example instantiate the `ArrayPostPolicy` instead.

Thanks a bunch for this gem - authorization never felt right to me until I found Pundit.

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -81,12 +81,7 @@ module Pundit
       elsif object.class.respond_to?(:policy_class)
         object.class.policy_class
       else
-        klass = if object.is_a?(Array)
-          object.map { |x| find_class_name(x) }.join("::")
-        else
-          find_class_name(object)
-        end
-        "#{klass}#{SUFFIX}"
+        "#{find_class_name(object)}#{SUFFIX}"
       end
     end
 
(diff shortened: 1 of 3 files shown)


delphaber:
+1 Happens to me too

Linuus:
Thank you for your contribution and sorry that no one has answered here earlier. 

By now this PR is quite old and I don't think it's appropriate to add Rails 3 specific things since it's EOL'd since a while back. Therefore I'll close this.

Thanks :heart:

PR #409, opened 2016-08-08
Add docs for closed system scopes

motine:
When I implemented a closed system, found myself using a scope before I call authenticate.
If you do this, the check if the user is logged in is done after the scope is evaluated. This requires the an explicit check for `user.nil?`. To avoid this, we can apply the same trick as suggested in the _Closed Systems_ section for scopes.

``` ruby
# controller example
  def index
    @apples = policy_scope(Apples)
    authorize @apples
  end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,7 +391,8 @@ rails g pundit:policy post
 
 In many applications, only logged in users are really able to do anything. If
 you're building such a system, it can be kind of cumbersome to check that the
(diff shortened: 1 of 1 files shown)


motine:
The CI failure is not due to this pull request. This pr only changes markdown files ;)

Linuus:
Thanks!

PR #418, opened 2016-09-08
Update broken README links to ensuring policies and scopes are used in pundit.rb

aaackerman:
Noticed  that the links to the "Ensuring Policies and Scopes are Used" section of the README were outdated in `pundit.rb`, this fixes the anchoring.


Linuus:
Thanks!

PR #419, opened 2016-09-27
Use controller’s `action_name` attr instead of params[:action]

Ross-Hunter:
This is the recommended way to access a controller’s current action -> http://guides.rubyonrails.org/action_controller_overview.html#routing-parameters

This change should reduce future maintenance if Rails deprecates accessing the action via `ActionController::Parameters`. It's also just a little cleaner.

It's also potentially more secure since we call `public_send` using the thing that get's pulled out of the params which is generally a bad idea. The 2 special keys inside of Rails' ActionControllerParams are `:controller` and `:action`, so it is not a really a vulnerability, but I think it's worth understanding the risk and moving away from the pattern.


Ross-Hunter:
There appears to be some issues with dependencies that are not tied to this code change. It looks like all PRs have been failing for a couple months now, even documentation changes. I will try to take a look at it if I get a chance.

mattvague:
Would also be nice to verify that there is in fact a current action. It's an edge case, but I'm currently unit testing a pundit-enabled controller (don't ask), and I was getting the following cryptic error because I hadn't manually set an `action` in my request:

`undefined method `?' for #<MyPolicy:0x007feb64d12a98>`.

Would be happy to submit a PR or help out however I can. Thanks for the great library!

Linuus:
Thanks!

PR #420, opened 2016-09-28
WIP Fix Bundle on CI

Ross-Hunter:
It looks like JRuby and Rubinius builds have been failing on CI for a couple months because bundler couldn't install some gems.

I think something might be up with currently with Travis, because now I am unable to install any version of Rubinius, even on [https://travis-ci.org/elabs/pundit/jobs/163589746](other branches) that didn't have this problem yesterday. I did get everything else to work by locking down some versions.

I'm going to try and rebuild it again tomorrow to see if the remaining issue is just an intermittent travis/rvm thing.


Ross-Hunter:
So MRI < 2.2 also seems to fail intermittently. Yikes.

Ross-Hunter:
I'm not able to resolve these CI issues.

PR #430, opened 2016-11-16
More accurate depiction of error raised

nathanl:
Show the actual error class and what the message would look like


jnicklas:
Nice, thanks!

PR #432, opened 2016-11-22
Add #authorize_with to allow specifying a policy class

nathanl:
For example, if both PostsController and PostModerationsController need
to check permission on Post, and different rules apply in each case,
PostModerationsController can call:

    authorize_with(record, policy_class: PostModerationPolicy)

I'm not sure if this is a feature you want, and if so, I'm happy to tweak the implementation and/or tests as you like.

Right now I have a use case where the same record is used in different controllers because the record is essentially a message: a `StaffingRequest` is created in the `StaffingRequestsController`, and in `StaffingRequestAssignmentsController`, different users assign a worker to that `StaffingRequest`.

I think you already support finding a different policy depending on namespace, but `Assignments::StaffingRequestsController` didn't seem like a good name, and it seemed pretty extreme to namespace my controller solely for the benefit of `authorize`.

My current workaround is to re-implement `authorize` in that controller, instantiating the policy I want, but it means I have to do stuff like `@_pundit_policy_authorized = true` myself, which seems pretty hacky.

So my options were:

1. Namespace my controller solely for the benefit of `authorize`
2. Re-implement `authorize` myself so that it can use the policy class I want
3. Add a version of `authorize` to Pundit that can be told what policy class to use

3 seemed like the best choice.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -176,17 +176,11 @@ protected
   # @raise [NotAuthorizedError] if the given query method returned false
   # @return [Object] Always returns the passed object record
   def authorize(record, query = nil)
-    query ||= params[:action].to_s + "?"
+    pundit_authorize_with_policy(record, query) {  policy(record) }
+  end
 
-    @_pundit_policy_authorized = true
-
-    policy = policy(record)
-
-    unless policy.public_send(query)
-      raise NotAuthorizedError, query: query, record: record, policy: policy
-    end
-
-    record
+  def authorize_with(record, policy_class:, query: nil)
+    pundit_authorize_with_policy(record, query) {  policy_class.new(pundit_user, record) }
   end
 
   # Allow this action not to perform authorization.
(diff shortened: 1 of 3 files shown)


lib/pundit.rb
```diff
@@ -271,6 +265,20 @@ def pundit_user
 
 private
 
+  def pundit_authorize_with_policy(record, query)
```
nathanl: Maybe not the best name? Mostly I'm trying to avoid any possible clashes with custom methods in the user's controller. I would just implement this on a different module/class, but it needs to set `@_pundit_policy_authorized`

nathanl:
Update: an option I (embarrassingly) didn't think of before, but a colleague just suggested, is to define `policy(record)` on my controller to return an instance of the policy I want to use.

```ruby
def policy(record)
  PostModerationPolicy.new(pundit_user, record)
end
```

I still think it's more straightforward if Pundit allows passing a policy in, but this is a pretty easy workaround.

nathanl:
Closing. Given that I can redefine `policy` in the controller to specify a policy class, I don't think it's worth Pundit having another feature to maintain.

PR #428, opened 2016-11-04
Fix build

jnicklas:

PR #435, opened 2016-12-06
Travis build change: Fix build, use latest JRuby

olleolleolle:
- this commit collects the changes made in #428

...while using the "let Bundler know which Ruby version we are currently using, so it can locate fitting gems on its own" `ruby` version annotation in the Gemfile.

This sweeping change:

- also moves all the dev dependencies needed into a Bundler group
- And ensures a modern-enough Bundler, so that the above method _works_.
- Works around a Rubocop version issue: lint only on a Ruby 2.4.1 with latest Rubocop


olleolleolle:
```

Error: obsolete parameter IndentWhenRelativeTo (for Style/CaseIndentation) found in /home/travis/build/elabs/pundit/.rubocop.yml
`IndentWhenRelativeTo` has been renamed to `EnforcedStyle`
obsolete parameter AlignWith (for Lint/EndAlignment) found in /home/travis/build/elabs/pundit/.rubocop.yml
`AlignWith` has been renamed to `EnforcedStyleAlignWith`
```

olleolleolle:
Hehe, I'll try putting the _old names_ of Cops in the .rubocop_todo.yml file, to make them apply.

olleolleolle:
@jnicklas Here! A Friday present: this should build green.

olleolleolle:
Oh, TODO: Do not fail Rubocop on 2.4.1.

jaredbeck:
Olle, if you're done working on this, I can pick up where you left off. However, I'm concerned that there aren't any maintainers in the discussion here. Can a maintainer please confirm that they will review a PR that fixes CI? CC: @jnicklas @thomasklemm

olleolleolle:
@jaredbeck Thank you for looking at this.

It made me lint the last bit and go green.

I think this is as ready is it will get. Can we get it merged?

ce07c3:
@jaredbeck @olleolleolle it's been abandoned more or less, maintainers are active twice a year. I'd suggest starting a new repo and releasing under a different gem name, e.g. pundit2.

thomasklemm:
Hey guys, wanted to post this week anyway. Sorry for letting this drag on a bit.

I'd like to hand over my spot as maintainer to someone else who is more active, since I'm too far away from the codebase to help at the moment. Will open another issue after talking with @jnicklas for finding a replacement maintainer for myself.

pundit.gemspec
```diff
@@ -19,12 +20,4 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
 
   gem.add_dependency "activesupport", ">= 3.0.0"
-  gem.add_development_dependency "activemodel", ">= 3.0.0"
```
thomasklemm: @olleolleolle This PR looks solid, thanks a lot.

Could you reiterate on these lines, a hint on why we can remove these development dependencies would be nice.

olleolleolle: @thomasklemm Oh, since all development takes place within the Bundler context, and we have better control of versions needed etc using Gemfiles, we can use a Group in Bundler (`:development`) and keep development-related things there.

ce07c3:
@thomasklemm thanks for responding, I'd be sad if this project were to become obsolete because it wasn't maintained. :)

pundit.gemspec
olleolleolle: Bundler's version-and-platform finding mechanism finds a gem version that suits the current running Ruby.

In order to give more background, here is an example.

The Bundler mechanism assists us when it comes to secondary deps: let's say a test tool that we explicitly refer to has a dependency. We get that pulled in implicitly. That dep needs to be "> 2" on Ruby 2+ and "< 2" on Ruby 1.9. If we're unlucky and have mentioned all this in the gemspec, we have to find and keep updated a least common denominator and that could be tricky.

thomasklemm: Cool, I see, thanks for the quick answer!

olleolleolle:
Thanks 😻

pundit.gemspec
dgmstuart: @olleolleolle I'm doing some work on the gem and trying to follow this thread. Can you explain how the ruby version example you mention above is any different when specifying gems in the `Gemfile`? My understanding is that a Gemfile can only specify one ruby version. 

Also, while we want the gem to be compatible with as many ruby versions as possible, for the purposes of _development_ of the gem I'm much more open to supporting a smaller set of ruby versions.

olleolleolle: Adding some notes, in order to clarify:

One of the key things is that a Ruby library repository often git-ignores its `Gemfile.lock`, so that that file is re-created at the `bundle install` invocation.

The single RUBY_VERSION becoming a single Ruby version number, and informs this CI run's creation of its `Gemfile.lock`.

dgmstuart: @olleolleolle Thanks for responding so quickly 😄  

The Bundler docs seem to recommend committing `Gemfile.lock`: presumably so that all developers of the gem are using the same versions: https://bundler.io/guides/creating_gem.html#testing-our-gem - I'm considering doing that.

Is the point you're making that somehow a lockfile can be different under different ruby versions? I don't know what the mechanism for that would be for that, but I can believe that it's true. What I still don't quite understand is why it would be any different with the gems specified in the Gemfile rather than the gemspec - sorry if I'm missing something obvious.

olleolleolle: The subtle note is: apps should def. check in their Gemfile.lock. Libraries, it's more subtle, there. Libraries want to support the widest possible matrix of Ruby versions.

- When we ask for `gem "todeloo", "~> 2.2"`, we're placing a constraint on the "solution" of the `Gemfile.lock` (i.e. its resulting content).
- When we ask for a specific `ruby 2.2.2`, we're placing _another_ constraint on the solution. [Docs section about this directive](https://bundler.io/v2.0/man/gemfile.5.html#RUBY)
- By writing `ruby RUBY_VERSION`, we make that constraint "the currently-running Ruby"

dgmstuart: @olleolleolle I totally agree that libraries in _usage_ want to support many ruby versions, but is the same true for libraries in _development_? Note that all the gems we're talking about here are development dependencies, so they should have no impact at all on _users_ of the gem, right? Only on _developers_ of the gem. 

Again, the docs that I linked to seem to be Bundler themselves specifically saying that `Gemfile.lock` should be checked in _for gems_, since it only impacts gem maintainers, not gem users. Or am I reading it wrong?

(Thanks for taking the time to go down this rabbithole with me by the way - I really appreciate it)

dgmstuart: I found a better link: https://bundler.io/man/bundle-install.1.html#THE-GEMFILE-LOCK

> When Bundler first shipped, the Gemfile.lock was included in the .gitignore file included with generated gems. Over time, however, it became clear that this practice forces the pain of broken dependencies onto new contributors, while leaving existing contributors potentially unaware of the problem. Since bundle install is usually the first step towards a contribution, the pain of broken dependencies would discourage new contributors from contributing. As a result, we have revised our guidance for gem authors to now recommend checking in the lock for gems.

olleolleolle: The rabbit-hole expands! Thanks for taking me to task and re-read the section.

_Recommended praxis has shifted & been documented!_

dgmstuart: @olleolleolle I'm not sure I understand your last comment?

olleolleolle: Thanks for telling me about the updated documentation! I wasn't aware.

PR #438, opened 2016-12-14
fix typo

orliesaurus:
fix typo :)


ce07c3:
https://en.wiktionary.org/wiki/scaleable

orliesaurus:
fair enough - just looked very odd to me :)

PR #441, opened 2016-12-19
Allow overriding policy class on authorize method

pablocrivella:
This PR allows to override the policy class on the ```authorize``` method. 
This is useful when you have a controller that manages different types of models that share a common interface and you want to use the same policy for that kind of behaviour.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -224,10 +224,11 @@ protected
   #
   # @see https://github.com/elabs/pundit#scopes
   # @param scope [Object] the object we're retrieving the policy scope for
+  # @param policy_scope_class [Class] the policy scope class we wan't to force use of
   # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
-  def policy_scope(scope)
+  def policy_scope(scope, policy_scope_class: nil)
     @_pundit_policy_scoped = true
-    pundit_policy_scope(scope)
+    policy_scope_class ? policy_scope_class.new(pundit_user, scope).resolve : pundit_policy_scope(scope)
   end
 
   # Retrieves the policy for the given record.
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -61,10 +61,11 @@ module Pundit
     # @param user [Object] the user that initiated the action
     # @param record [Object] the object we're checking permissions of
     # @param query [Symbol, String] the predicate method to check on the policy (e.g. `:show?`)
+    # @param policy_class [Class] the policy class we want to force use of
     # @raise [NotAuthorizedError] if the given query method returned false
     # @return [Object] Always returns the passed object record
-    def authorize(user, record, query)
-      policy = policy!(user, record)
+    def authorize(user, record, query, policy_class: nil)
+      policy = policy_class ? policy_class.new(user, record) : policy!(user, record)
 
       raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
@@ -189,7 +190,7 @@ protected
   # @param record [Object] the object we're checking permissions of
   # @param query [Symbol, String] the predicate method to check on the policy (e.g. `:show?`).
   #   If omitted then this defaults to the Rails controller action name.
-  # @param policy_class [Class] the policy class we wan't to force use of
+  # @param policy_class [Class] the policy class we want to force use of
   # @raise [NotAuthorizedError] if the given query method returned false
   # @return [Object] Always returns the passed object record
   def authorize(record, query = nil, policy_class: nil)
@@ -224,7 +225,7 @@ protected
   #
   # @see https://github.com/elabs/pundit#scopes
   # @param scope [Object] the object we're retrieving the policy scope for
-  # @param policy_scope_class [Class] the policy scope class we wan't to force use of
+  # @param policy_scope_class [Class] the policy scope class we want to force use of
   # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
   def policy_scope(scope, policy_scope_class: nil)
     @_pundit_policy_scoped = true
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -189,14 +189,15 @@ protected
   # @param record [Object] the object we're checking permissions of
   # @param query [Symbol, String] the predicate method to check on the policy (e.g. `:show?`).
   #   If omitted then this defaults to the Rails controller action name.
+  # @param policy_class [Class] the policy class we wan't to force use of
   # @raise [NotAuthorizedError] if the given query method returned false
   # @return [Object] Always returns the passed object record
-  def authorize(record, query = nil)
+  def authorize(record, query = nil, policy_class: nil)
     query ||= "#{action_name}?"
 
     @_pundit_policy_authorized = true
 
-    policy = policy(record)
+    policy = policy_class ? policy_class.new(pundit_user, record) : policy(record)
 
     raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
--- a/README.md
+++ b/README.md
@@ -133,6 +133,18 @@ def publish
 end
 ```
 
+You can pass an argument to override the policy class if necessary. For example:
+
+```ruby
+def create
+  @publication = find_publication # assume this method returns any model that behaves like a publication
+  # @publication.class => Post
+  authorize @publication, policy_class: PublicationPolicy
+  @publication.publish!
+  redirect_to @publication
+end
+```
+
 If you don't have an instance for the first argument to `authorize`, then you can pass
 the class. For example:
 
--- a/README.md
+++ b/README.md
@@ -283,6 +283,15 @@ def show
 end
 ```
 
+Like with the authorize method, you can also override the policy scope class:
+
+``` ruby
+def index
+  # publication_class => Post
+  @publications = policy_scope(publication_class, policy_scope_class: PublicationPolicy::Scope)
+end
+```
+
 Just as with your policy, this will automatically infer that you want to use
 the `PostPolicy::Scope` class, it will instantiate this class and call
 `resolve` on the instance. In this case it is a shortcut for doing:
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -375,6 +375,10 @@ describe Pundit do
       expect { controller.authorize(post, :destroy?) }.to raise_error(Pundit::NotAuthorizedError)
     end
 
+    it "can be given a different policy class" do
+      expect(controller.authorize(post, :create?, policy_class: PublicationPolicy)).to be_truthy
+    end
+
     it "works with anonymous class policies" do
       expect(controller.authorize(article_tag, :show?)).to be_truthy
       expect { controller.authorize(article_tag, :destroy?) }.to raise_error(Pundit::NotAuthorizedError)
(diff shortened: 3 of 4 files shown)


prschmid:
Hey @pablocrivella can you also include this functionality for the ``policy_scope`` function so that it can be used to resolve #77

pablocrivella:
Hey @prschmid, will do!

prschmid:
@pablocrivella 
image

pablocrivella:
@prschmid could you take a look?

prschmid:
@pablocrivella just what I hoped it to be. =) 

I was literally searching for a solution to this very problem this morning, was looking for a way to solve it/get around it, and then stumbled across your PR. So, thank you very much!

pablocrivella:
@prschmid np, i'm glad i can contribute, now lets hope it gets approved and merged :)

prschmid:
Me too! Right now my solution is the ugly/potentially dangerous combination of skipping a policy and then manually calling the right one... something along the lines of
```ruby
skip_policy_scope
@objs = MyCustomPolicy::Scope.new(current_user, @resource_class)
```

ce07c3:
@prschmid, or workaround that is less unsafe (albeit a bit hackyish-meta):

```ruby
    def policy_classify(target,
                                   klass)
      target.define_singleton_method(:policy_class) do
        klass
      end

      target
    end
```

prschmid:
@ce07c3 I'll give that a shot! Thanks!

pablocrivella:
@elabs-dev do you know how could i solve the failing tests?

prschmid:
@pablocrivella Looking at one of the results from travis-ci (https://travis-ci.org/elabs/pundit/jobs/185173933), it looks to be an issue with Rake. Looks like you're not the only one. Hope this helps.
http://stackoverflow.com/questions/35893584/nomethoderror-undefined-method-last-comment-after-upgrading-to-rake-11

Fahrenhei7:
Does someone still work on it?

pablocrivella:
@Fahrenhei7 what do you mean? i'm still waiting for a collaborator to merge it :)

Linuus:
@pablocrivella 

Hi!

Sorry for the delayed responses here.

Can't you just manually specify the policy class in your model/object?

See here: https://github.com/elabs/pundit#manually-specifying-policy-classes

```ruby
class Post
  def self.policy_class
    PublicationPolicy
  end
end

class Article
  def self.policy_class
    PublicationPolicy
  end
end
```

pablocrivella:
Hey @Linuus, the problem with that is it will be always overriding the policy_class. 
In my case, i have models that have some specific concerns like ```publishable``` i.e. and i have a separate controller that manages all this ```publishable``` models. 
That is the only place where i would like to override the policy class.

Linuus:
Ah, I see.

I am a bit concerned that this is quite a big edge case that not many users would ever need, but maybe I'm wrong? :)

It can be easily done in your concern though if you override the `authorize` method

```ruby
  # Add this to the concern

    skip_authorization

    query ||= params[:action].to_s + "?"

    policy = PublicationPolicy.new(current_user, record)

    unless policy.public_send(query)
      raise NotAuthorizedError, query: query, record: record, policy: policy
    end

    record
  end
```

pablocrivella:
@Linuus oh i get that. That is actually what i'm doing but on my application controller, and since  i have found myself in this situation several times i thought about doing a PR 😅

Fahrenhei7:
exactly the same thing for me. :)

pablocrivella:
@Linuus maybe there is a cleaner way to override it without copying most of the authorize method?

Fahrenhei7:
@pablocrivella 
I did just like that:
```ruby
raise Pundit::NotAuthorizedError unless PersonPolicy.new(current_client, @person).show?
```

Linuus:
@Fahrenhei7 There are some problems with that version though. For instance it would not work together with `verify_authorized`.


@pablocrivella Ok, unless anyone else objects I'll merge this. Before that, could you add the same functionality to the class methods `authorize` and `policy_scope` please?

Here:
https://github.com/pablocrivella/pundit/blob/036b56d6dbe6d25c06cb78d91b677cb87f814718/lib/pundit.rb#L62-L81

ce07c3:
I still don't get the use case, though - why not simply create a new model and pass that to Pundit?

pablocrivella:
@ce07c3 I use it like in the following scenario:

```ruby
# app/models
class BlogPost < ApplicationRecord
  include Followable
end

class Conversation < ApplicationRecord
  include Followable
end

# app/controllers
class FollowablesController < ApplicationController
  before_action :authenticate_user!

  respond_to :js

  def create
    load_followable
    authorize(@followable, policy_class: FollowablePolicy)
    current_user.profile.follow(@followable)
    render :toggle
  end

  def destroy
    load_followable
    authorize(@followable, policy_class: FollowablePolicy)
    current_user.profile.unfollow(@followable)
    render :toggle
  end

  private

  def load_followable
    @followable ||= followable_klass.find(params[:id])
  end

  def followable_type
    @followable_type ||= params[:type]
  end

  def followable_klass
    followable_type.classify.constantize
  end
end

# app/policies
class FollowablePolicy < ApplicationPolicy
  def create?
    return false unless user.editor? || user.member?
    return false if record.reportable? && record.hidden?
    !record.followed_by?(user)
  end

  def destroy?
    return false if record.reportable? && record.hidden?
    record.followed_by?(user)
  end
end
```

ce07c3:
I'd suggest to whitelist the followable_type, otherwise an attacker/creative user might be inclined to pass in `User` or other class strings you wouldn't expect.

I understand the use case but I would just wrap the followable into a `FollowableWrapper` and then use a `FollowableWrapperPolicy` (substitute "Wrapper" with whatever makes most sense to you, naming wise). That way, you can continue using Pundit's class resolver.

pablocrivella:
@ce07c3 You are right about the whitelist, thats on my todo list.

About the wrapper, sounds good also, i guess i could use something like SimpleDelegator.
But, IMHO i like more this explicit approach, since is pretty clear that we are overriding the policy class just for this specific scenario.
But i get your point and i thank you for showing me other approaches :)
If this is too much of an edge case, it might not be good to merge this PR.

I'm okey with whichever decision is made in order to keep the code clean and understandable.

Linuus:
@ce07c3 Thank you for your input. Using something like `SimpleDelegator` would be a nice way of solving this actually.

@pablocrivella Thank you for your contribution, but if it's OK I'm going to close this for now and if other people request this feature in the future we can reopen and merge this.

Sounds good?

❤️ 💚 💜

pablocrivella:
Sure @Linuus np :) let me know if i could contribute in order ways!

bbugh:
I was so excited to finally find this pull request after searching for ways to specify the policy class, and I am disappointed to see that it wasn't ever merged. I've had to hack around this in so many projects! How many people do you need asking for this feature before you'll merge it? 😄

pablocrivella:
@bbugh You can override this methods on you application controller, that is what i have been doing :)

pablocrivella:
@bbugh maybe you could add the scenarios where you needed this?

i.e. recently i had to use this with a Parent -> Childs models scenario, sending the parent as the `record` for the child policy class.

jondayton:
Ran into this and would upvote. Scenario was also a Parent -> Child models

pablocrivella:
@jondayton yeah i still use this code as a patch in most of my projects 😅

wynksaiddestroy:
I would love to get this merged too! 👌🏾

Linuus:
I’m currently on parental leave but I will take a look at this again when I get back!

rkrdo:
Any updates on this PR? Seems like it handles just what I currently need

Linuus:
@rkrdo It's on my todo list. I haven't gotten to it yet though.

pablocrivella:
@Linuus Updated this PR latest changes on master

Linuus:
I've been thinking about this a bit... Another approach to solve this would be to use namespaces and inheritance, right?

Example:

```ruby
class BlogPost < ApplicationRecord
  include Followable
end

class Conversation < ApplicationRecord
  include Followable
end

# app/controllers
class FollowablesController < ApplicationController
  def create
    load_followable
    authorize([:publishable, @followable])
  end
end

# app/policies
module Publishable
  class FollowablePolicy < ApplicationPolicy
    def create?
      return false unless user.editor? || user.member?
      return false if record.reportable? && record.hidden?
      !record.followed_by?(user)
    end
  end

  class BlogPost < FollowablePolicy; end
  class Conversation < FollowablePolicy; end
end

```

Then it wouldn't be necessary to override the policy manually, you don't have to wrap the objects and you still have explicit policies for each domain model where you can put special logic if need be.

What do you think about this?

pablocrivella:
@Linuus That looks like a really nice solution for this specific scenario! 
I still think it would be really useful to have a way to explicitly set the desired policy.

But i understand your point of that there are better/cleaner ways to achieve the same result.

I like how [actionpolicy](https://actionpolicy.evilmartians.io/#/quick_start?id=basic-usage) does this using a `with` option on the `authorize!` method.

ce07c3:
Explicitly passing a policy might make sense, as the authorize method does some other lighter lifting such as ensuring flags are set for the verify filters. It offers a very explicit and convenient way of addressing the issue of resolving policies in a straightforward manner.

Now, I know that Pundit is mostly about resolving policies automatically - that's the main point of the gem, together with a few other principles (clean Policy POROs). I just don't see this would be such a bad addition nevertheless, and we could still offer namespace based resolution. :)

Linuus:
@ce07c3 

> Explicitly passing a policy might make sense, as the authorize method does some other lighter lifting such as ensuring flags are set for the verify filters

What do you mean? In my example the flags are set? :)

I'm not saying this is a big or bad change, I'm just trying to figure out if this is a good pattern or if there is a better approach which just needs some documentation to be solved.

jensljungblad:
How would these additional context arguments be passed to the Policy? As the `user` argument or as a new argument?

Linuus:
The context stuff was just an example and really doesn’t have anything to do with this PR :) We can discuss that in a separate issue.

lib/pundit.rb
```diff
@@ -189,14 +189,15 @@ def verify_policy_scoped
   # @param record [Object] the object we're checking permissions of
   # @param query [Symbol, String] the predicate method to check on the policy (e.g. `:show?`).
   #   If omitted then this defaults to the Rails controller action name.
+  # @param policy_class [Class] the policy class we wan't to force use of
```
Linuus: wan't => want

lib/pundit.rb
```diff
@@ -223,10 +224,11 @@ def skip_policy_scope
   #
   # @see https://github.com/elabs/pundit#scopes
   # @param scope [Object] the object we're retrieving the policy scope for
+  # @param policy_scope_class [Class] the policy scope class we wan't to force use of
```
Linuus: wan't => want

Linuus:
Random thought: Should we pass `policy_class` to the `policy_scope` method as well and infer the `::Scope` part? Hm, perhaps it's better to just being explicit though... after all it is an override feature.

ce07c3:
@Linuus explicit 👍 

Regarding your previous answer, I think I was referring to some bugs I had seen in the past. Likely jumped around with my thoughts a bit. Your example: just use a wrapper that includes both records. I wouldn't allow passing in extra context.

Linuus:
Yeah it was just something I’ve thought about.

pablocrivella:
@Linuus updated

Linuus:
Great, thank you! Unfortunately the JRuby builds are failing but it seems to be an issue with Travis. See https://github.com/travis-ci/travis-ci/issues/9826

It’s probably fixed soon though! I’m keeping an eye on the Travis issue.

pablocrivella:
@Linuus i think the issue is now fixed

PR #442, opened 2016-12-23
Add a little clarity and direction for testing scopes

atomanyih:
Wasted a little time earlier looking for the recommended way to test scopes. I think most people know how to test ruby objects, but might assume that there is a different way to do it because of the nice Policy spec DSL. At least that's what I did!

Just a line like this would make it ever so slightly clearer and might have saved me a few minutes

Seems to have come up a few times, so maybe I'm not the only one: 
https://github.com/elabs/pundit/issues/333
https://github.com/elabs/pundit/pull/137
https://github.com/elabs/pundit/issues/123


Linuus:
Hm, I want to avoid too much noise in the readme... But OK, I'll merge it :)

Thanks

Issue #368, opened 2016-02-26
Getting a scope for a specific action


AndrewSwerlick:
Is there a mechanism for having a policy with different scopes for different kinds of actions. For example, I have a license table that stores all the licenses for our application that an organization has purchased. When users login, they have a page where they can select a license to operate under. Company administrators have pages where they can view a license, and all its associated details. "select" vs "view" are two very different permissions. Users may be able to select licenses from their entire company, but may only be responsible for the administration of licenses in their department, and so can only view those. However in both cases I have reasons to want to quickly create a scope of the records the user is authorized to access. For "select" I need to populate drop downs, for "view", I need to build out tables on index pages. I don't see how I can do this in pundit with the current scope setup.

nbekirov:
> Is there a mechanism for having a policy with different scopes for different kinds of actions.

If I'm getting this right maybe you'll be interested in this discussion: [Allows Scope to receive additional arguments](https://github.com/elabs/pundit/pull/10)

AndrewSwerlick:
@nbekirov I have seen that, but it seemed a little different in goals, even if the implementation ends up being similar. From the example included, it looks like that user it trying to pass in somewhat arbitrary arguments, and for that reason @jnicklas suggests wrapping in a domain object. I'm interested only in passing additional "verbs" into scope, generally the same "verbs" that correspond to the policy's method names.

davidstosik:
:+1: 

It would be nice to have different scopes for different actions, such as:

- Users I'm allowed to see
- Users I'm allowed to manage (me only, unless I'm admin)
- Users I'm allowed to mail (all but me if I'm admin, only my friends otherwise)

Would there be any red flag in implementing multiple Scopes in the same Policy?

```ruby
# user_policy.rb
class UserPolicy < ApplicationPolicy
  class Scope < Scope  # equivalent to `< ApplicationPolicy::Scope`
    def resolve
      if user.admin?
        scope
      else
        scope.confirmed.enabled  # filter unconfirmed, and disabled accounts
      end
    end
  end

  class EmailScope < Scope  # equivalent to `< User::Scope`
    def resolve
      super.where.not(users: {id: user.id})  # Based on the UserPolicy's main Scope
    end
  end
end

# in my views/controllers
# ...
mailable_users = UserPolicy::EmailScope.new(current_user, User).resolve
# ...
```

jnicklas:
Yes, I think that's a nice pattern. Nothing wrong with that in my book! :+1:

[...]

PR #443, opened 2016-12-29
Allows additional methods to be used with Scopes

davidbiehl:
A proposal for issue #368.

Adds an additional argument to the `policy_scope` methods that specifies the method on the scope that should be called. Defaults to `:resolve` for backward compatibility.

This makes Scopes more flexible by allowing multiple scopes to be defined on a `Scope` class and making them available for use in controllers or views with existing helper methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,45 @@ You can, and are encouraged to, use this method in views:
 <% end %>
 ```
 
+It is also possible to define additional methods the scope that you can use in
+different situations. We'll add an `unpublished` scope to the 
+`PostPolicy::Scope`:
+
+``` ruby
+class PostPolicy < ApplicationPolicy
+  class Scope < Scope
+    def resolve
+      if user.admin?
+        scope.all
+      else
+        scope.where(published: true)
+      end
+    end
+
+    def unpublished
+      if user.admin?
+        scope.all
+      else
+        scope.where(published: false)
+      end
+    end
+  end
+
+  def update?
+    user.admin? or not post.published?
+  end
+end
+```
+
+To use the `unpublished` scope, simply pass the name of the method as the 2nd 
+argument to `policy_scope`:
+
+``` ruby
+def index
+  @posts = policy_scope(Post, :unpublished)
+end
+```
+
 ## Ensuring policies and scopes are used
 
 When you are developing an application with Pundit it can be easy to forget to
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -74,10 +74,11 @@ module Pundit
     # @see https://github.com/elabs/pundit#scopes
     # @param user [Object] the user that initiated the action
     # @param scope [Object] the object we're retrieving the policy scope for
+    # @param method [Symbol] the method to call on the scope. Defaults to :resolve.
     # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
-    def policy_scope(user, scope)
+    def policy_scope(user, scope, method = :resolve)
       policy_scope = PolicyFinder.new(scope).scope
-      policy_scope.new(user, scope).resolve if policy_scope
+      policy_scope.new(user, scope).public_send(method) if policy_scope
     end
 
     # Retrieves the policy scope for the given record.
@@ -85,10 +86,11 @@ module Pundit
     # @see https://github.com/elabs/pundit#scopes
     # @param user [Object] the user that initiated the action
     # @param scope [Object] the object we're retrieving the policy scope for
+    # @param method [Symbol] the method to call on the scope. Defaults to :resolve.
     # @raise [NotDefinedError] if the policy scope cannot be found
     # @return [Scope{#resolve}] instance of scope class which can resolve to a scope
-    def policy_scope!(user, scope)
-      PolicyFinder.new(scope).scope!.new(user, scope).resolve
+    def policy_scope!(user, scope, method = :resolve)
+      PolicyFinder.new(scope).scope!.new(user, scope).public_send(method)
     end
 
     # Retrieves the policy for the given record.
@@ -116,8 +118,8 @@ module Pundit
 
   # @api private
   module Helper
-    def policy_scope(scope)
-      pundit_policy_scope(scope)
+    def policy_scope(scope, method = :resolve)
+      pundit_policy_scope(scope, method)
     end
   end
 
@@ -209,10 +211,11 @@ protected
   #
   # @see https://github.com/elabs/pundit#scopes
   # @param scope [Object] the object we're retrieving the policy scope for
+  # @param method [Symbol] the method to call on the scope. Defaults to :resolve.
   # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
-  def policy_scope(scope)
+  def policy_scope(scope, method = :resolve)
     @_pundit_policy_scoped = true
-    pundit_policy_scope(scope)
+    pundit_policy_scope(scope, method)
   end
 
   # Retrieves the policy for the given record.
@@ -271,7 +274,10 @@ protected
 
 private
 
-  def pundit_policy_scope(scope)
-    policy_scopes[scope] ||= Pundit.policy_scope!(pundit_user, scope)
+  def pundit_policy_scope(scope, method = :resolve)
+    method = method.to_sym
+    key    = method == :resolve ? scope : [scope, method]
+
+    policy_scopes[key] ||= Pundit.policy_scope!(pundit_user, scope, method)
   end
 end
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -43,6 +43,10 @@ describe Pundit do
       expect(Pundit.policy_scope(user, Post)).to eq :published
     end
 
+    it "returns an instantiated policy scope given the method to call on the scope" do
+      expect(Pundit.policy_scope(user, Post, :unpublished)).to eq :unpublished
+    end
+
     it "returns an instantiated policy scope given an active model class" do
       expect(Pundit.policy_scope(user, Comment)).to eq Comment
     end
@@ -67,6 +71,10 @@ describe Pundit do
   describe ".policy_scope!" do
     it "returns an instantiated policy scope given a plain model class" do
       expect(Pundit.policy_scope!(user, Post)).to eq :published
+    end
+
+    it "returns an instantiated policy scope given the method to call on the scope" do
+      expect(Pundit.policy_scope!(user, Post, :unpublished)).to eq :unpublished
     end
 
     it "returns an instantiated policy scope given an active model class" do
@@ -409,6 +417,10 @@ describe Pundit do
   describe "#policy_scope" do
     it "returns an instantiated policy scope" do
       expect(controller.policy_scope(Post)).to eq :published
+    end
+
+    it "returns an instantiated policy scope when a scope method is provided" do
+      expect(controller.policy_scope(Post, :unpublished)).to eq :unpublished
     end
 
     it "throws an exception if the given policy can't be found" do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -30,6 +30,10 @@ class PostPolicy < Struct.new(:user, :post)
     def resolve
       scope.published
     end
+
+    def unpublished
+      scope.unpublished
+    end
   end
 
   def update?
@@ -60,6 +64,10 @@ end
 class Post < Struct.new(:user)
   def self.published
     :published
+  end
+
+  def self.unpublished
+    :unpublished
   end
 
   def to_s


ce07c3:
Nice addition. Might wish to call it `scope` instead of `method`?

davidbiehl:
The methods I'm adding the argument to already have an argument called `scope` ... I'm not a huge fan of `method` but it seemed to be the most straightforward name considering scope was already taken. Do you have any other suggestions?

davidbiehl:
Maybe `resolver`? It might be too semantically similar to the default `:resolve` tho.

ce07c3:
@davidbiehl I have little hope that this will ever be merged so I am forking and working from there.

davidbiehl:
For anyone that stumbles on this, you can also just include the following method in your `ApplicationController`. This does completely override the default helper method, so proceed with caution.

```ruby
def policy_scope(scope, method = :resolve)
  policy_scope = PolicyFinder.new(scope).scope
  policy_scope.new(current_user, scope).public_send(method) if policy_scope
end
```

masciugo:
@davidbiehl thanks for your workaround. I had to verride `pundit_policy_scope` as it consider the usage of `verify_policy_scoped`. See https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L210-L213

Linuus:
@davidbiehl Sorry for this PR being left alone for soo long.

I'm a bit curious if you have some more specific use cases for this?

My first thought is that Pundit should do the base scoping and not cater for different views. Like, if I'm allowed to see my organization's articles then that's the base scope. Whether I only can see published articles in some views and unpublished in another is not Pundit's responsibility (I'm open to be convinced about this :) ).

I mean, something like this seems appropriate for me:

```ruby
# In ArticlePolicy
def resolve
  scope.where(organization_id: user.organization_id)
end

# In FooController
policy_scope(Article).published

# In BarController
policy_scope(Article).unpublished

# In BazController
policy_scope(Article)
```

I don't see the benefit of passing another argument which controls the extra scoping for different scenarios.

What do you think?

immerrr:
@Linuus consider the following example.

> DISCLAIMER: I have only recently started working with Rails and Pundit was shown to me literally less than an hour ago, so my code might have odds and quirks, but should be enough to convey the idea.

Suppose in your example a user can have roles assigned on per-article basis, e.g. 'READER', 'EDITOR', 'AUTHOR' and so on. It is handled by an `ArticleUser` many-to-many association with a `role` attribute. A scopeless policy would look something like this:

```
class ArticlePolicy
  def show?
    user.admin? || article.article_users.any?(user: user)
  end

  def update?
    user.admin? || article.article_users.find_by?(user: user, role: ['EDITOR', 'AUTHOR']) 
  end

  def delete?
    user.admin? || article.article_users.find_by?(user: user, role: ['AUTHOR']) 
  end
end
```
Then the "base scope" would be something like
```
  class Scope
    attr_reader :user, :articles

    def initialize(user, articles)
      @user  = user
      @articles = articles
    end

    def resolve
      if user.admin?
        articles.all
      else
        articles.where('EXISTS(SELECT 1 FROM article_users WHERE article_id = articles.id AND user_id = ?)', user.id)
      end
    end
  end
```

With that base scope, 
```
readable_articles = policy_scope(Article)
```
will return a list of readable articles, no problem. 

The problem is that to go from `readable_articles` to `deletable_articles` for non-admin users you need to refer to `article_users` table again, but now filtering by `AUTHOR` role. It can be remedied somewhat by doing an `INNER JOIN` instead of `EXISTS` in `resolve` method and then doing
```
readable_articles.where('article_users.role = ?', 'AUTHOR')
```
but that could be more expensive performance-wise and that would mean we would also have to pay the JOIN performance penalty for the base scope. And you also need to take care if a user can have multiple roles for an article, because then you need to deduplicate the scope too.

Now, if only at the moment of `resolve`'s execution we would know what action is expected to be run on the objects in scope, we could do the filtering right away with the second scope method:
```
    def resolve_for_delete
      if user.admin?
        articles.all
      else
        articles.where('EXISTS(SELECT 1 FROM article_users WHERE article_id = articles.id AND user_id = ? AND role = ?)', user.id, 'AUTHOR')
      end
    end
```
And avoid any performance penalties or deduplication problems.

Linuus:
@immerrr why use a scope for the delete action? You can just authorize that specific article in the `destroy?` method as you do in your example?

immerrr:
> @immerrr why use a scope for the delete action? You can just authorize that specific article in the `destroy?` method as you do in your example?

For example, I'm implementing a multi-choice form to batch-delete articles, then I want a scope to be able to display the list of articles that a user can destroy and maybe paginate through it. 

And I mean, I can check objects one by one with `destroy?` method, but it's the same thing as the "basic" scope: you can go through a Rails relation and authorize objects one by one for `show?` too, it's just that it is wildly inefficient.

vcavallo:
@Linuus 

> I'm a bit curious if you have some more specific use cases for this?
> I don't see the benefit of passing another argument which controls the extra scoping for different scenarios.
> 
> What do you think?

Here's a potential simple example: date range params passed in from the request.

So you want to limit the results based on the date range first and foremost and then apply some Pundit-controlled policy restrictions as well - but every user type will definitely be scoped down by those date params. You'd have to apply a sort of "where dates in" query to the model first (in the controller) and then pass that subset of results into the Pundit scope. If I understand correctly, that would be (at least one) additional separate query, whereas if those params could be passed along into Pundit it could possibly all be handled in one larger query.

Incidentally, this happens to be exactly what I'm trying to do and how I found this issue :)

dmitry:
What about add scopes similarly to the https://actionpolicy.evilmartians.io/#/scoping ?
Scopes are not just about activerecord relation scopes, but also could be about filtering request params, or even mutating an attribute data of the model, like the name of the user, which could be abbreviated in case when other users might not see his real full name.

Linuus:
@vcavallo 
> If I understand correctly, that would be (at least one) additional separate query, whereas if those params could be passed along into Pundit it could possibly all be handled in one larger query.

No. Well, it depends on your ORM, but most ORMs are lazy so there will only be one query executed in the end.
Also, what you want is to make it possible to pass additional params to the scopes, which is not what this PR is about.

@dmitry 
> What about add scopes similarly to the https://actionpolicy.evilmartians.io/#/scoping ?
Scopes are not just about activerecord relation scopes, but also could be about filtering request params, or even mutating an attribute data of the model, like the name of the user, which could be abbreviated in case when other users might not see his real full name.

IMO this is too complex of a feature to be in Pundit. It kind of goes against the core philosophy of the library. However, I don't think I'm the lead maintainer anymore so ultimately I guess it's up to @dgmstuart to decide.

dgmstuart:
I'm just starting to get my head around this and I have a couple of small thoughts for starters:

- I believe naming a method `method` overrides [`Object#method`](https://ruby-doc.org/core-2.6.3/Object.html#method-i-method), which can lead to some pretty funky situations. I've been burned by that before
-  I'm still not clear on what sort of use case there is for having a single `Scope` class with two methods on it, rather two `Scope` classes with one method. If there are two scopes which share some logic then I wonder if a better solution would be to extract that logic to a third class? 

The fact that we're passing class names rather than instances perhaps makes that less clean than it might otherwise be (eg. I don't see how we could have one `Scope` decorating another), but that's perhaps a bigger discussion.

dgmstuart:
Ah wait, the suggestion is around having a _variable_ called `method` not a method.

dgmstuart:
Possibly dumb question:

For @immerrr's example does the following make any sense? What are the tradeoffs?
```ruby
 class ReadScope
    attr_reader :user, :articles

    def initialize(user, articles)
      @user  = user
      @articles = articles
    end

    def resolve
      if user.admin?
        articles.all
      else
        articles.where('EXISTS(SELECT 1 FROM article_users WHERE article_id = articles.id AND user_id = ?)', user.id)
      end
    end
  end
```
```ruby
 class DeleteScope
    attr_reader :user, :articles

    def initialize(user, articles)
      @user  = user
      @articles = articles
    end

    def resolve
      if user.admin?
        articles.all
      else
        articles.where('EXISTS(SELECT 1 FROM article_users WHERE article_id = articles.id AND user_id = ? AND role = ?)', user.id, 'AUTHOR')
      end
    end
  end
```
```ruby
readable_articles = policy_scope(Article, policy_scope_class: ArticlePolicy::ReadScope)
deletable_articles = policy_scope(Article, policy_scope_class: ArticlePolicy::DeleteScope)
```

...although there seems to be some mismatch between the README which includes the example:

```ruby
@publications = policy_scope(publication_class, policy_scope_class: PublicationPolicy::Scope)
```
...and the actual source which says:
```ruby
 def policy_scope(user, scope)
```
(https://github.com/varvet/pundit/blob/5b0d709144e020139f7440d035cf7ac9975c19ae/lib/pundit.rb#L84)

...so maybe this doesn't exist anymore 😞

immerrr:
It's not stupid at all, it could work, although there's an immediate code duplication optimization, the
```ruby
   attr_reader :user, :articles

    def initialize(user, articles)
      @user  = user
      @articles = articles
    end
```
part is asking to be shared between the policy classes. Or if it is just one class, to be shared between methods with no extra actions :)

dgmstuart:
@immerrr is there more duplication between these two classes than between two `Scope` classes relating to different resources? I'm not sure there is, but I'm ready to be convinced.

immerrr:
I guess it depends on the policy itself, I mean the policy in question obviously shares this snippet between two resolve methods:
```
      if user.admin?
        articles.all
      else
```
But none of this is a big deal really.

The way you emphasize "different" resources could be the reason why we went for different approaches intuitively. I think it's because I don't see them as different resources, the resource for me here is the same, the list of objects passed in as the input scope. Then the policies look like filters rather than new subresources to me. And the policy scopes are a natural optimization of what otherwise would look something like
```
policy = Policy.new(...)
scope.select { |obj| policy.show?(obj) }
# or 
scope.select { |obj| policy.update?(obj) }
```

But I don't insist. For me personally, it would be great to see this feature land, the *way* it is implemented doesn't matter that much.

dgmstuart:
@immerrr I think I've caused some confusion: when I say different resources I do literally mean different - eg. `UserPolicy::Scope` vs `ArticlePolicy::Scope`.

Wouldn't this bit be duplicated regardless of whether this was implemented in the same class or two different classes?
```ruby 
if user.admin?
  articles.all
 else
```
According to the documentation, the solution I outlined (which is actually the same as this, described in the original issue: https://github.com/varvet/pundit/issues/368#issuecomment-196111115), should already be possible in pundit (but whether the documentation is correct is another matter).

immerrr:
Ah, ok, sorry.

> Wouldn't this bit be duplicated regardless of whether this was implemented in the same class or two different classes?

Originally I had it in my head as something like

```ruby
def resolve_for_action(action: :show?)
  if user.admin?
    articles.all
  else
    ...
  end
end 
```

And in that case there would probably less duplication, but like I said, I'm fine with separate methods or separate classes too, you would split the single method if/when it would grow anyway.

Re: the solution you outlined: I get how from the maintainer's perspective sometimes less features is better than more features, both in terms of effort and interface complexity. As the example from #368 shows, your approach can be achieved without any pundit interference at all:
```ruby
  @publications = policy_scope(publication_class, policy_scope_class: PublicationPolicy::Scope)
  # with
  @publications = PublicationPolicy::Scope.new(current_user, publication_class).resolve
```

That approach does not require any extra pundit code, and oddly enough it ends up being a few characters shorter.

AFAICS this PR is more about a way to bring feature parity between `authorize` and `policy_scope` in terms of the interface, to get them closer to something like:
```ruby
authorize post, :update?
# vs
policy_scope Post, :update?
```
The way it is implemented IMHO doesn't really matter as long as the implementation details, like policy class names, [don't leak](https://en.wikipedia.org/wiki/Leaky_abstraction).

Linuus:
> ```ruby
> readable_articles = policy_scope(Article, policy_scope_class: ArticlePolicy::ReadScope)
> deletable_articles = policy_scope(Article, policy_scope_class: ArticlePolicy::DeleteScope)
> ```
> 
> ...although there seems to be some mismatch between the README which includes the example:
> 
> ```ruby
> @publications = policy_scope(publication_class, policy_scope_class: PublicationPolicy::Scope)
> ```
> 
> ...and the actual source which says:
> 
> ```ruby
>  def policy_scope(user, scope)
> ```
> 
> (
> 
> https://github.com/varvet/pundit/blob/5b0d709144e020139f7440d035cf7ac9975c19ae/lib/pundit.rb#L84
> )
> 
> ...so maybe this doesn't exist anymore 😞

That's not the correct method you are looking at. Here is the source:

https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L250-L253

dgmstuart:
Ah I see - there's a class method and an instance method with the same name.

Linuus:
> Ah I see - there's a class method and an instance method with the same name.

Yes. So you can do `Pundit.policy_scope(....)` in other places if you want. See https://github.com/varvet/pundit/issues/537 for more info why overriding the policy class is not available on the class method.

dgmstuart:
OK - I think I'm getting there with understanding this - thanks everyone for your patience. Here's my understanding of the issue:

1. In principle a Pundit `Policy` can have different rules for each of the 7 RESTful controller actions [i], as well as arbitrary other methods
2. By default these policy rules get applied in the corresponding controller action.
3. For _some_ of the RESTful controller actions [ii], we may want to define which _list_ of existing records that action can act on for a given user. This is what `Scope` is intending to do.
4. In addition: if we have our own non-RESTful policy rules we may also want to specify the list of records for which those rules are true 
5. In general there is a relationship between a policy rule and the corresponding scope: 

    - the **_policy rule_** asks a true/false question of a single user+record combination
    - the **_scope_** should return the list of things for which that _same question_ is true 

6. In many cases, these lists will be identical for all policy rules [iii]. This is why the current approach (which supports a single scope) is sufficient for many use cases.
7. Unlike policy rules, scopes will not generally be applied in the corresponding controller action [iv]

**Conclusion 1:** Different scopes corresponding to different policy rules is a well-formed concept. Let's refer to these as 'scope rules'

**Conclusion 2:** Scope rules should _only_ relate to specific policy rules (otherwise there's no reason for them to live within Pundit rather than in the model)

**Conclusion 3:** The behaviour of `policy_scope` isn't really the same shape as `authorize`, for two reasons: 
- `authorize post, :create?` is meaningful, but `policy_scope Post, :create?` is not a well-formed concept ("the list of `Posts` which can be created")
- `authorize post` has default behaviour based on the context: it looks up the corresponding policy rule for the current controller action, which defaults to false. With `policy_scope Post` we can't in general infer anything from the context, so the default behaviour is something we have to define.

I wonder if it would be misleading to aim to give these such similar interfaces since they don't actually behave the same way? (cc @immerrr)

**Conclusion 4:** Since a specific scope rule and policy rule are actually two representations of the same rule, ideally we would infer the scope rule from the policy rule rather than defining it explicitly, but I don't see how that's possible since pundit rules are arbitrary ruby and we can't assume much about what's backing the policy scope either.

Given some data we could perhaps check that the policy and scope rules match up, but I'm not sure if that's a useful/practical thing to do: 

```ruby
post = policy_scope(Posts, :updateable).first
authorize(post, :update?) # should always be true
```

--- 

- [i] ...though in practice it's hard to imagine a case for `new`/`create` and `edit`/`update` having different rules
- [ii] If we assume that the RESTful actions correspond to records backed by a database then this is probably `index`, `show`, `edit`, `update` and `destroy` since they're actions which relate to existing records. However it's hard to imagine that the scope for `edit` would be different to `update`.
- [iii] Example: Blog Posts.
  - everyone can see published things
  - users can see, edit, and delete all posts which they have authored
  - admins can see and do everything that any user can do. 
- [iv] Example: We might have a screen showing a list of editable records (`index`) in which we use `policy_scope Record, :editable`. Each record in the list has a button which toggles some value on that record (`update`), but there is no dedicated page for modifying individual records (no `edit` action)

immerrr:
Great summary, thanks @dgmstuart ! Sorry for the delay.

> authorize post, :create? is meaningful, but policy_scope Post, :create? is not a well-formed concept ("the list of Posts which can be created")

Yeah, the "new"/"create" pair (C of CRUD) is a strange thing to do for a Scope that's supposed to operate on persisted objects, but the rest should fit quite well.

> authorize post has default behaviour based on the context: it looks up the corresponding policy rule for the current controller action, which defaults to false. With policy_scope Post we can't in general infer anything from the context, so the default behaviour is something we have to define.

In that sense `authorize` and `policy_scope` methods are not exactly equivalent and might not need the same API. It would be nice to make the two APIs close at least in the part where the operation is specified explicitly.

> Since a specific scope rule and policy rule are actually two representations of the same rule, ideally we would infer the scope rule from the policy rule rather than defining it explicitly, but I don't see how that's possible since pundit rules are arbitrary ruby and we can't assume much about what's backing the policy scope either

Perhaps it could it be done the other way around? E.g. auto-generating a policy rule from a scope rule by doing something like `Scope.resolve(Post.where(id: post.id)).exists?` at a cost of an extra query? Or maybe there's a library to provide query methods over a collection of pre-fetched model instances so that `Scope.resolve([post]).exists?` would work...

Happy Holidays!

dgmstuart:
@immerrr thanks for your response

> It would be nice to make the two APIs close at least in the part where the operation is specified explicitly.

What's the reasoning here? In what ways is this nice? My concern is that this would be taking two things which are actually quite different, and setting up the expectation that they're the same, which could lead to some problems.

> Perhaps it could it be done the other way around? E.g. auto-generating a policy rule from a scope rule

That's a nice idea, but I'm pretty sure the problems go both ways: bear in mind that policies and scopes can both involve arbitrary code referring to the user, as well as the model.

Eric-Guo:
I need this feature, my situation is when do a reporting, one report should allow user see the underground data, so I have to loose the scope, another report is now allow, so using the normal current pundit scope. this PR resolve my problem.

So is it possible to merge it? (I can resolve the conflict and propose a new PR if needed.)

dgmstuart:
Hi @Eric-Guo it's been a long time since I've had my head in this, so bear with me. 

I'm not sure I understand your use case - what do you mean by "underground data"?

My main question is: why do you need these scopes to be part of Pundit? Are they actually to do with authorization, or could they just be scopes on the model in question?

I don't think I'm OK with merging this PR in its current form: at least the given example doesn't feel meaningful to me:
```ruby
    def unpublished
      if user.admin?
        scope.all
      else
        scope.where(published: false)
      end
    end
```

How should we interpret this method? If it's "the list of unpublished posts", that's not correct, since if the user is an admin it includes published posts.

Pundit is centered around verbs: "what can users _do_?". The base use case of policy scopes is "what is the list of records that the user can _see_?". "unpublished" doesn't fall into that category.

If you have a need to show a list of unpublished posts in your application, for me that's a concern of your model, not of Pundit: 

```ruby
# model:
scope :unpublished, -> { where(published: false) }

# controller:
Posts.unpublished
``` 

If there's some additional restriction on that then the approach is to use the policy scope:
```ruby
policy_scope(Post).unpublished
```

If you can describe your use case more specifically, maybe I can help.

dmitry:
> why do you need these scopes to be part of Pundit
Application could have different contexts.

dgmstuart:
@dmitry I'm not sure I understand what you mean. Can you give an example?

Note that 
```ruby
policy_scope(Post.unpublished)
```

Also seems to be a possibility.

Eric-Guo:
Here is [my usage](https://github.com/thape-cn/cybros_portal/commit/d91f700d3289ae1ceebe00b32a05bde8197ed26b#diff-3c315d6eba922e16201a06f319db25b0R5), I having two different report which fetch data from same table, The user can be category in 5 different:

``` ruby
  ALL_OF_ALL = 1
  ALL_EXCEPT_OTHER_COMPANY_DETAILS = 2
  MY_COMPANY_ALL_DETAILS = 3
  MY_COMPANY_EXCEPT_OTHER_DEPTS = 4
  MY_DEPARTMENT = 5
```

One of the report need to show the summary of the group (group having many different subsidiary) and another report need show summary of the subsidiary only.

If user belongs to ALL_EXCEPT_OTHER_COMPANY_DETAILS, it means such user can see the both report.

My case here is I must allow ALL_EXCEPT_OTHER_COMPANY_DETAILS see the both group report and her/his subsidiary report, but I can not allow ALL_EXCEPT_OTHER_COMPANY_DETAILS user see other company reports.

So for the policy class of the Active Record SubCompanyNeedReceive, there is two different scope need to resolve. one is general and another is group resolve.

dgmstuart:
@Eric-Guo I'm sorry, I'm afraid I'm finding it difficult to understand you.

The code you linked to is also extremely complicated - would it be possible for you to invent a simpler example?
Could you create an example involving two scopes and show how both scopes would be called in controllers?

Or one thing you could try to explain that might help me understand:
In `app/policies/bi/sub_company_need_receive_policy.rb` there are two scopes, but only one method (`show?`). 

Does this mean that users with the `User::ALL_EXCEPT_OTHER_COMPANY_DETAILS` code are allowed to see records (?) which aren't listed in the default scope?

Eric-Guo:
> Does this mean that users with the User::ALL_EXCEPT_OTHER_COMPANY_DETAILS code are allowed to see records (?) which aren't listed in the default scope?

Yes, that's the key question, the User::ALL_EXCEPT_OTHER_COMPANY_DETAILS, not allow see the record which listed in the group_resolve, that's why those record filtered in the default resolve, but User::ALL_EXCEPT_OTHER_COMPANY_DETAILS need to see the summary result of filtered record.

I can indeed fetch all data using `Bi::SubCompanyRealReceive`. but I have 5 different type of user so without this PR, I need to put the such user control logic in the [controller](https://github.com/thape-cn/cybros_portal/blob/thape/app/controllers/report/subsidiary_receives_controller.rb#L23-L57), so I need this PR to put such logic return to `SubCompanyRealReceivePolicy`.

I have to say the business logic is some what strange, they tell me such kind of user, ALL_EXCEPT_OTHER_COMPANY_DETAILS, can see the summary data, but can not see the detail data which in fact, the calculating source of the summary data. from the scientific perspective, really make no sense, anyway the world is still driving by business, so I have to say this requirement is a really business require...

dgmstuart:
@Eric-Guo I think I'm asking a different question. I'm afraid your actual use case is probably too complicated for me to realistically understand the situation. Can you create a simpler example?

Basically I agree with Linus: it seems like the idea of multiple scopes is to do _filtering_, rather than actual authorization, and so shouldn't be something that Pundit gets involved in

I'm very open to being convinced otherwise, but I need to see a realistic use case that I can understand.

Eric-Guo:
>  the idea of multiple scopes is to do filtering, rather than actual authorization

Yes, agree, the result of authorization is do filtering in tables, the row authorization in particular tables really need put in the corresponding [user writing pundit policy](https://github.com/thape-cn/cybros_portal/blob/thape/app/policies/bi/contract_hold_sign_detail_policy.rb#L7-L16) class, that's really the great design of gem pundit!

My problem is the report I writes in Rails, need show user the summary of the `contract_hold` table result, but need to stop user see other department `contract_hold` [details](https://github.com/thape-cn/cybros_portal/blob/thape/app/policies/bi/contract_hold_policy.rb#L21).

I guess maybe my reporting is only the case which need this feature, so I would quite understanding it's not quite good to let this PR merge. Waiting to see if other people can give more particular real example in future, for me, I would like thank @davidbiehl because this PR really save me many time.

dgmstuart:
@Eric-Guo

> Yes, agree, the result of authorization is do filtering in tables,

No, that's not what I'm saying. 

"filtering" is an ambiguous word. This is what I mean when I'm using that word:
There are two different concepts here: 
1. **Authorization:** Listing _all_ the records that a user is allowed to interact with in _any_ context
2. **Filtering:** Listing a _subset_ of the records that a user is allowed to interact with, in a _specific_ context

Pundit is [designed to help with 1](https://github.com/varvet/pundit#scopes):
> Often, you will want to have some kind of view listing records which a particular user has access to.

Pundit should probably _not_ get involved in 2: 
Instead you can use the policy scope (Pundit) in combination with regular scopes (Rails) on your models.

To put it another way, the intention is that the definition of `resolve` should list exactly the set of records for which `show?` is true. That doesn't seem to be the case in [your code](https://github.com/thape-cn/cybros_portal/blob/d91f700d3289ae1ceebe00b32a05bde8197ed26b/app/policies/bi/sub_company_need_receive_policy.rb).

Eric-Guo:
Thank your kindly reply and patient to explain your design decision.

I'm afraid I still hope `Pundit` can help rails developer resolve case 2. because in case 2, in a **specific** context, the control variable is probably only *model class* and *user*, if I can not put such logic in policy class. I only can allow User::ALL_EXCEPT_OTHER_COMPANY_DETAILS access **all** record in every usage and do a rails regular scopes in every usage place at same time, which more repeat and no a prevent mistakes design.

So this PR, IMHO merged to the master will be helpful. 😄

dmitry:
Why not to add filtering to the pundit as it would be helpful to filter-out unauthorized attributes posted via HTTP using strong_parameters?

dgmstuart:
@Eric-Guo 

> because in case 2, in a specific context, the control variable is probably only model class and user

I'm not sure I understand what you mean:

Let's say we have records A, B, C and D, and we have a user who is authorised in general to see A, B and C.

Let's say we have a page on which we only want the user to be able to see A and B. They're authorised to see C, but we don't want to include it on this page for some other reason.

How can that reason be _only_ based on the user? Don't we need to use some information about the records in order to decide which ones to show?

Eric-Guo:
> How can that reason be only based on the user? Don't we need to use some information about the records in order to decide which ones to show?

We always control access based on user and his access tables. we do need some information about the user, like his title, job_level, company or departments, but it can be calculated on the [User#calculate_operation_access_code](https://github.com/thape-cn/cybros_portal/blob/thape/app/models/user.rb#L161)

I need change the example a little bit: we have records A, B, C and D in the one table, and we have a user who is authorised in *general* to see A, B only and it used in many place

Now we having a little special page, he should be able to see A,B and C. (notice he still can not access D.)

Then can not using the general resolve in pundit because C is already filled out in general, and if we not using `policy_scope` in special page, it works, but have to put the user control logic in controller and if another special page coming, it need to repeat in that controller again. (maybe using rails scope can help minimal the logic, but still need to repeat using that scope again), let it even make it more worse, if user having different access level, then such logic will become very [long](https://github.com/thape-cn/cybros_portal/blob/thape/app/policies/bi/contract_hold_policy.rb#L19-L30).

That's the one reason I want to merge this PR, another reason is some what optioned, pundit as `authorization gem` should help developer put all their authorization logic in the policy class, although it will probably will change the meaning of pundit slogan (Minimal authorization through OO design and pure Ruby classes)

dgmstuart:
@Eric-Guo If they're authorized to see record C in _any_ context, then _by definition_ they're authorized to see it.

If you want to hide that record in some contexts then that's not authorization and it's not something Pundit should be concerned with. 

Your concept of "authorized in general" doesn't make sense to me - sorry.

dgmstuart:
Thanks for your input everyone, but I'm closing this PR. 

I'll very happily re-open if someone is able to present (or construct) a simple example of a plausible use case that I can understand.

dmitry:
@dgmstuart please look to the activepolicy scoping examples: https://actionpolicy.evilmartians.io/#/scoping

dgmstuart:
@dmitry thanks, I've read them. Currently the reason for closing is that I haven't seen a use case that I can understand, for which this feature seems like a good solution. I don't see one in those docs either: it's all pretty abstract no?

All I'm asking for is a simple plausible example where two different scopes required in the same policy.

All I can say for certain about the actionpolicy feature is that it seems pretty complex, and as Linus said above, against the design approach of this gem. Bear in mind that asking for new features is asking for me to be part of maintaining those features for the foreseeable future.

dmitry:
@dgmstuart few examples:

- You may filter out parameters using scopes for the `strong_parameters`, and not just filtering (based on the user role) records using AR (or any other ORM).
- Sometimes there are different contexts to fetch the data (with filter based on the user role), for example you may have two types of documents in the user model (ID and business license). Both documents have one class (`Document`), but a bit different context. Means you have to differentiate between them.

dgmstuart:
@dmitry I'm sorry I'm afraid I don't understand the first point. Could you write a simple code example?

In the second case, it sounds like ID and business license are treated differently for the purposes of authorization, in which case shouldn't they have different policies? Or if I've misunderstood, please write a simple code example: it's much easier to understand by looking at code.

dmitry:
### First case.

Strong parameters scope usage could look like:

```ruby
policy_scope(Post, :params)
```

Definition:
```ruby
 class ParamsScope
    def initialize(user, posts)
      @user  = user
      @posts  = posts
    end

    def resolve
      if user.admin?
        [:user_id, :title, :text]
      else
        [:text]
      end
    end
  end
```

---

### Second case

I agree you can create shared class (DocumentPolicy) which can be reused as nested class for IDPolicy and BusinessLicensePolicy classes. One of the real examples, is when you are logged in as a super user and you want to control over subuser's documents. In that case you need to have special context scope, as you are not accessing directly from the user's account to the it's documents, but from the super account.

dgmstuart:
@dmitry 
First case: 
That's not a scope though: a scope returns a list of records.
If you want to be able to support arbitrary functions like this which return different results based on the user role, that's a whole other discussion.

dgmstuart:
@dmitry Second case sounds completely implementable in Pundit currently. I'm showing an example with different policies, but I don't think you actually need these: it seems much simpler to just pass the relevant scope into the `policy_scope` method.

``` ruby
class Admin::DocumentsController < ApplicationController
  def index
    @documents = policy_scope(Document)
  end
end

class DocumentsController < ApplicationController
  def index
    @documents = policy_scope(Document)
  end
end

class IDCardsController < ApplicationController
  def index
    @documents = policy_scope(Document, policy_scope_class: IDCardPolicy::Scope)
    # but actually this isn't even necessary and you can just do this for exactly the same result:
    @documents = policy_scope(Document.id_cards)
  end
end

class DocumentPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      if user.admin?
        scope.all
      else
        scope.where(owner_id: user.id)
      end
    end
  end
  
  def show?
    record.type = "licence" || user.admin? || record.owner_id = user.id
  end
end

class LicencePolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      scope.all
    end
  end
  
  def show?
    true
  end
end

class IDCardPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      if user.admin?
        scope.id_cards # assuming that the Document model has a scope which returns all the ID cards
      else
        scope.id_cards.where(owner_id: user.id)
      end
    end
  end
  
  def show?
    user.admin? || record.owner_id = user.id
  end
end
```

PR #446, opened 2017-01-09
Custom policy finder

eperezks:
This pr allows users to support multiple workflows for their application.  For example, let's say that your app has the concept of projects.  Not just one project but multiple with different authorization and scoping.  In this case you might have a couple of rails engines, ProjectWorkflowOne and ProjectWorkflowTwo that are namespaced with the engine name but leverage the same common models.  By using the new PrefixPolicyFinder you'll be able to separate the policy/scoping logic for each workflow into their files.  

Set it for the entire engine within the the application.rb file.
before_action { |_p| set_pundit_prefix 'ProjectWorkflowOne::' }

Note that it doesn't matter if you are using engines or not.  You can specify a 'prefix' for you policies thereby allowing you to easily use namespaced policies.


Linuus:
I think this can be solved by the namespaces feature, right?
See: https://github.com/varvet/pundit#policy-namespacing

PR #447, opened 2017-01-12
Only rescue NameError related to wanted policy constants

faucct:
Hello. When you are rescuing all `NameError`'s, you are also rescuing the errors not related to the constant you are trying to get - the ones that happen during autoload, for example. This will result in user getting wrong error messages:

```ruby
# autoloadable.rb
class AutoloadablePolicy < Missing
end
```
`Pundit.policy!(user, Autoloadable)` will throw ``Pundit::NotDefinedError: unable to find policy `AutoloadablePolicy` for `user` `` instead of `NameError: uninitialized constant Missing`.
Another example:
```ruby
# autoloadable.rb
class AutoloadablePolicy
  foo
end
```
`Pundit.policy!(user, Autoloadable)` will throw the same error instead of ``undefined local variable or method `foo' for main:Object``.
 This can confuse the gem user.

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -23,9 +23,7 @@ module Pundit
     #   scope.resolve #=> <#ActiveRecord::Relation ...>
     #
     def scope
-      policy::Scope if policy
-    rescue NameError
-      nil
+      "#{policy}::Scope".safe_constantize if policy
     end
 
     # @return [nil, Class] policy class with query methods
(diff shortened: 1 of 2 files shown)


thomasklemm:
@faucct Thank for this PR. Could you rebase on `master` since Travis is now green again?

Linuus:
Fixed by #491

faucct:
Sorry for not rebasing. I guess I have missed the notification :(

PR #461, opened 2017-05-24
Fix code examples in Scopes section

iiska:
Rename reference to `` post `` to `` record `` which would be currently correct one when PostPolicy class inherits ApplicationPolicy.


Linuus:
Thanks! 👍

PR #473, opened 2017-07-16
Refactor policy generator

yhirano55:
I've refactored policy generators:

- Changed generator namespace from `rails g pundit:policy something` to `rails g policy something`. Because I think that Policy is custom layer for Rails application same as Decorator ([active_decorator](https://github.com/amatsuda/active_decorator#usage), [draper](https://github.com/drapergem/draper#generators)). How about it?
- Refactored source_root
- chmod -x
- Updated `README`

Please let me know if you have advice or questions. Thanks a lot!


yhirano55:
I really appreciate your code-review. Thanks.

dgmstuart:
@yhirano55 I agree with @damien - there's some good stuff here but maybe there are two PRs required? 

One PR could contain the refactorings (e.g. the source_route change) - that seems like an improvement and easy to merge.

The other PR could contain the suggested namespace change: this isn't a refactoring, since it changes the behaviour of the gem. 

I can imagine that the gem maintainers might want to keep the `Pundit` namespace: this keeps the code separate from anything provided by core Rails: Imagine that in someone's project they define their own custom generator called `PolicyGenerator` - or even that Rails decides to implement one: we don't want Pundit's functionality to override that or be overriden by it. Having the namespace ensures that the Pundit code is handled separately.

Either way: having a PR specifically for that change would make a place for that discussion to happen.

Linuus:
@yhirano55 Thank you for your contribution!

I agree with @dgmstuart that this should be split in two PRs. I'm 👍  on the `source_route` refactoring but not sure on the namespace change. I like that it's clear that the generator is related to pundit.

If you make it a separate PR we can discuss it there though.

I'll close this and look forward to a new PR (or two ;) ).

yhirano55:
Sorry... I've forgotten this patch. I'll create new PR. I appreciate your comments. Thanks.

PR #475, opened 2017-07-22
fix PolicyFinder when given an array with policy class override

ryanhertz:
When trying to find a policy using an array, the `policy_class` override method was not getting called on the last item of the array.

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -36,7 +36,7 @@ module Pundit
     #   policy.update? #=> false
     #
     def policy
-      klass = find
+      klass = find(object)
       klass = klass.constantize if klass.is_a?(String)
       klass
     rescue NameError
@@ -48,7 +48,7 @@ module Pundit
     #
     def scope!
       raise NotDefinedError, "unable to find policy scope of nil" if object.nil?
-      scope or raise NotDefinedError, "unable to find scope `#{find}::Scope` for `#{object.inspect}`"
+      scope or raise NotDefinedError, "unable to find scope `#{find(object)}::Scope` for `#{object.inspect}`"
     end
 
     # @return [Class] policy class with query methods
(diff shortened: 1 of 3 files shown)


ryanhertz:
@damien @jnicklas Can this be merged? It's my first real OS pr.

mikaellisselgard:
What the status on this one? Policies are currently broken when trying to use the namespace functionality used in #331. Anything I can do?

armchairdj:
Please merge this! I have now forked Pundit twice, once last year and once just now after pulling upstream changes from the last few months, to fix this same issue. And my fix was not nearly as elegant! Kudos to @ryanhertz for tackling it!

Linuus:
This looks good to me. Thanks @ryanhertz !

halo:
Just to mention it for posterity. I use [Draper](https://github.com/drapergem/draper) and occasionally try to authorize a `Draper::CollectionDecorator` like so:

```ruby
def index
  @articles = authorize scope.decorate
end
```

For whatever reason, the `Draper::CollectionDecorator` masks [as an Array](https://github.com/drapergem/draper/blob/master/lib/draper/collection_decorator.rb#L63-L66). Since this change, Pundit tries to find a Policy with the name "Article::Article::Article::ArticlePolicy", depending on the number of records in the Array 😆 

I know, I shouldn't be using Draper at all, but I thought I'd just mention the conflict here.

May I also ask how this pull request is related to https://github.com/varvet/pundit/issues/351 ? Do they interfere with each other? Does this one fix the other one? Or are they not related? I'm just trying to make sense out of the code.

Thank you for your hard work!

Linuus:
@halo #351 is not really related to this one. The problem here is that if you override the `policy_class` method on the object being authorized it wouldn't call it if used with the namespacing functionality, e.g. `authorize [:admin, @article]`.

We still always pass the entire array to the policy in the end. I'm not sure if that will change yet.

Issue #482, opened 2017-10-02
Use policy_scope with custom class


laertispappas:
Hello I would like your feedback on something that I am implementing. 
I have explicitly required all actions in my application controller to authorize all requests:

```ruby
class ApplicationController < ActionController::Base
  include Pundit
  after_action :verify_authorized
end
```
Given I have a `posts controller` with an `index` action I would like to fetch all posts that are published for normal users. So the following is implemented in `PostsController` and `PostPolicy`:

```ruby
class PostsContoller
  def index
     posts = policy_scope(Post)
     render json: posts
   end
end

class PostPolicy < ApplicationPolicy
  class Scope < Scope
     def resolve
        scope.published
     end
  end
end
```   

What I would like to do is to create and `AdminPollPolicy` that will be used from an `Admin::PostsController` in order to fetch all `posts` regardless of their published status, thus I have the following:

```ruby
  class Admin::PostsController
    def index
      # posts = policy_scope(Post) # <== this will not work since it will call PostPolicy instead
     AdminPostPolicy::Scope.new(current_user, Poll).resolve
    end
  end

class AdminPostPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      user.admin? ? scope.all ? scope.none
    end
  end
end
```

However the above will raise a `PolicyScopingNotPerformedError` since we bypassed `policy_scope` method that mutates the corresponding instance variable in order to pass PolicyScope authorization:

```ruby

def policy_scope(scope)
    @_pundit_policy_scoped = true
    pundit_policy_scope(scope)
  end
```

Am I missing something here? What would be the correct way to implement the above?
Thanks.

laertispappas:
Maybe this is a duplicate of #480

ingemar:
👀

K-Eo:
You can use a namespace Policy:
```ruby
policy_scope([:admin, Post])
```
Check this PR #392.

laertispappas:
@K-Eo Thank you for your feedback. It looks like that is the way to go. Thanks.

PR #392, opened 2016-05-07
Add documentation for namespace support; resolves #373

daveriess:

PR #487, opened 2017-10-07
Resolves the correct scope for namespaced scopes 

laertispappas:
What I was trying to do is addressed here #482. The documentation for namespace support is addressed here #392. In summary a bug exist when we call `policy_scope` for a namespace resource.

Previously when we called `scope_policy([:admin, Post])`, the correct `Admin::PostPolicy::Scope` class is found but `scope` in `Admin::PostPolicy::Scope{@scope = [:admin, Post]}` holds the passed in array instead of `Post` class. 

This PR fixes the above issue to take into account the last argument of the passed in array as the scope if the passed in object responds to `:each`.

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -1,4 +1,46 @@
 module Pundit
+  class ScopeResolver
+    attr_reader :policy
+    attr_reader :user
+    attr_reader :resource_scope
+
+    def initialize(user, resource_scope)
+      @user = user
+      @resource_scope = resource_scope.respond_to?(:each) ? resource_scope.last : resource_scope
+      @policy = PolicyFinder.new(resource_scope).policy
+    end
+
+    def resolve
+      scope_class.new(user, resource_scope).resolve if scope_class
+    end
+
+    def resolve!
+      scope_class!.new(user, resource_scope).resolve
+    end
+
+  private
+
+    # @return [nil, Scope{#resolve}] scope class which can resolve to a scope
+    # @see https://github.com/elabs/pundit#scopes
+    # @example
+    #   scope = finder.scope #=> UserPolicy::Scope
+    #   scope.resolve #=> <#ActiveRecord::Relation ...>
+    #
+    def scope_class
+      policy::Scope if policy
+    rescue NameError
+      nil
+    end
+
+    # @return [Scope{#resolve}] scope class which can resolve to a scope
+    # @raise [NotDefinedError] if scope could not be determined
+    #
+    def scope_class!
+      raise NotDefinedError, "unable to find policy scope of nil" if resource_scope.nil?
+      scope_class or raise NotDefinedError, "unable to find scope for `#{policy}` policy"
+    end
+  end
+
   # Finds policy and scope classes for given object.
   # @api public
   # @example
@@ -16,18 +58,6 @@ module Pundit
       @object = object
     end
 
-    # @return [nil, Scope{#resolve}] scope class which can resolve to a scope
-    # @see https://github.com/elabs/pundit#scopes
-    # @example
-    #   scope = finder.scope #=> UserPolicy::Scope
-    #   scope.resolve #=> <#ActiveRecord::Relation ...>
-    #
-    def scope
-      policy::Scope if policy
-    rescue NameError
-      nil
-    end
-
     # @return [nil, Class] policy class with query methods
     # @see https://github.com/elabs/pundit#policies
     # @example
@@ -41,14 +71,6 @@ module Pundit
       klass
     rescue NameError
       nil
-    end
-
-    # @return [Scope{#resolve}] scope class which can resolve to a scope
-    # @raise [NotDefinedError] if scope could not be determined
-    #
-    def scope!
-      raise NotDefinedError, "unable to find policy scope of nil" if object.nil?
-      scope or raise NotDefinedError, "unable to find scope `#{find}::Scope` for `#{object.inspect}`"
     end
 
     # @return [Class] policy class with query methods
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -1,58 +1,15 @@
 module Pundit
-  class ScopeResolver
-    attr_reader :policy
-    attr_reader :user
-    attr_reader :resource_scope
-
-    def initialize(user, resource_scope)
-      @user = user
-      @resource_scope = resource_scope.respond_to?(:each) ? resource_scope.last : resource_scope
-      @policy = PolicyFinder.new(resource_scope).policy
-    end
-
-    def resolve
-      scope_class.new(user, resource_scope).resolve if scope_class
-    end
-
-    def resolve!
-      scope_class!.new(user, resource_scope).resolve
-    end
-
-  private
-
-    # @return [nil, Scope{#resolve}] scope class which can resolve to a scope
-    # @see https://github.com/elabs/pundit#scopes
-    # @example
-    #   scope = finder.scope #=> UserPolicy::Scope
-    #   scope.resolve #=> <#ActiveRecord::Relation ...>
-    #
-    def scope_class
-      policy::Scope if policy
-    rescue NameError
-      nil
-    end
-
-    # @return [Scope{#resolve}] scope class which can resolve to a scope
-    # @raise [NotDefinedError] if scope could not be determined
-    #
-    def scope_class!
-      raise NotDefinedError, "unable to find policy scope of nil" if resource_scope.nil?
-      scope_class or raise NotDefinedError, "unable to find scope for `#{policy}` policy"
-    end
-  end
-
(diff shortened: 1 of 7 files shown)


timcheadle:
Just ran into this and came up with the same solution. 👍

lib/pundit/policy_finder.rb
```diff
@@ -1,4 +1,46 @@
 module Pundit
+  class ScopeResolver
```
dgmstuart: It feels like a great improvement to extract a class to do this work, particularly since it separates out a responsibility from policy finder.

Since it's non-trivial, might there be value in testing its methods directly? Are there different combinations where the values of `resource_scope` and `policy` are various different types, which might be worth testing?

Also since this is a class we're calling directly, does it belong in its own named file?

lib/pundit/policy_finder.rb
```diff
@@ -1,4 +1,46 @@
...
+      scope_class!.new(user, resource_scope).resolve
+    end
+
+  private
+
+    # @return [nil, Scope{#resolve}] scope class which can resolve to a scope
+    # @see https://github.com/elabs/pundit#scopes
+    # @example
+    #   scope = finder.scope #=> UserPolicy::Scope
+    #   scope.resolve #=> <#ActiveRecord::Relation ...>
```
dgmstuart: Does this comment still make sense in this new context? Do private methods need doc comments?

lib/pundit/policy_finder.rb
```diff
@@ -1,4 +1,46 @@
...
+    #
+    def scope_class
+      policy::Scope if policy
+    rescue NameError
+      nil
+    end
+
+    # @return [Scope{#resolve}] scope class which can resolve to a scope
+    # @raise [NotDefinedError] if scope could not be determined
+    #
```
dgmstuart: Is this comment still required now that this is a private method?

lib/pundit/policy_finder.rb
```diff
@@ -1,4 +1,46 @@
...
+
+  private
+
+    # @return [nil, Scope{#resolve}] scope class which can resolve to a scope
+    # @see https://github.com/elabs/pundit#scopes
+    # @example
+    #   scope = finder.scope #=> UserPolicy::Scope
+    #   scope.resolve #=> <#ActiveRecord::Relation ...>
+    #
+    def scope_class
```
dgmstuart: Perhaps this would be a refactoring too far, but could this be even simpler if it was a question could ask the scope itself? e.g. by passing the resource scope into some sort of decorator which exposes `class` and `class!` methods.

lib/pundit/policy_finder.rb
laertispappas: No it doesn't. Will remove them thanks.

lib/pundit/policy_finder.rb
laertispappas: I don't think that adding a decorator at this point will give any benefits. Nice thought though.

lib/pundit/policy_finder.rb
laertispappas: You're points are correct. Methods should be directly tested and this is the main reason I din't extract the class into a separate file. The assumptions behind this class is that it should be tested "directly" from `pundit_spec`. `PolicyFinder` does not have any unit tests as well so I kept the same behavior here. 

I agree to extract the class to a separate file. Let me gather all your comments and try to make some changes. Thanks.

dgmstuart: I actually think it would be fine to test through `pundit_spec` if you wanted to go that way -
- I think it's totally fine to have nested classes, particularly when they are unstable or where it's not clear where the seam should lie, since this makes them easier to change. The trade-off is often slightly more complicated or obtuse tests.

The thing is that in this case I'm not sure that all of the possible configurations of ScopeResolver are currently exercised by that test.

Probably easier to test in a separate file anyway 👍

lib/pundit/policy_finder.rb
```diff
@@ -1,4 +1,46 @@
 module Pundit
+  class ScopeResolver
+    attr_reader :policy
+    attr_reader :user
+    attr_reader :resource_scope
+
+    def initialize(user, resource_scope)
+      @user = user
+      @resource_scope = resource_scope.respond_to?(:each) ? resource_scope.last : resource_scope
```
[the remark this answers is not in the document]
dgmstuart: It depends: if you want to accept specifically both Arrays and single values then this approach is possible (Inspired by Avdi Grimm's "Confident Ruby"):

```ruby
Array(resource_scope).last
```

If this needs to support arbitrary `Enumerable`s (which is what your code currently does) then I'm not sure. I don't really understand the surrounding context of this code. 

The fact that we're having to accept data which might be of different shapes feels like a code smell, but I'm not sure we can get away from that, since the rest of the code is full of `respond_to?`.

laertispappas:
@dgmstuart I have made some changes. Check out my last commit. Do you have any idea why the build is  failing on 2.4.1? (https://travis-ci.org/elabs/pundit/jobs/300633838)

dgmstuart:
@laertispappas Good stuff - I'll have a look.

Regarding the Travis failure, if I run 2.4.1 locally I get the same failures>

The issue is that the `.rubocop.yml` file specifies `TargetRubyVersion: 1.9` but the latest rubocop (0.51.0) has dropped support for that. 

Note that the last time `master` was built (6 months ago), the rubocop version was 0.48.0, which is why the repo shows the build as passing: https://travis-ci.org/elabs/pundit/jobs/231147775 

There's no restriction on the rubocop version in the Gemfile, which means that effectively any PR happening after a new rubocop version first needs to update the code to support the latest version of Rubocop.

The earliest version which Travis builds is 2.1, so I'd recommend updating the `.rubocop.yml` to specify that as the target version, which should fix this.

spec/spec_helper.rb
```diff
@@ -25,6 +25,12 @@ module PunditSpecHelper
   config.include PunditSpecHelper
 end
 
+class NonePolicy
+end
+
+class None
+end
```
dgmstuart: These are only used in that one spec, so it's a shame to put them in the spec helper, but that does seem to be the approach for other tests. Would an alternate approach be to define these in a module?

lib/pundit/policy_finder.rb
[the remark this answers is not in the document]
dgmstuart: @laertispappas what did you decide about this?

spec/scope_resolver_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+module Pundit
+  describe ScopeResolver do
+    subject { described_class.new(user, resource_scope) }
+    let(:user) { double('pundit_user') }
+
+    describe '#resolve' do
+      context 'policy class cannot be found' do
+        let(:resource_scope) { nil }
+
+        it { expect(subject.resolve).to be nil }
```
dgmstuart: This one-liner syntax works best when making statements directly about the subject (rather than methods called on the subject), which allows use of `is_expected` (see [docs](https://relishapp.com/rspec/rspec-core/docs/subject/one-liner-syntax)).

As it is, the subject feels like a bit of a [mystery guest](https://robots.thoughtbot.com/mystery-guest).

Would these tests be clearer if the `subject` and `before` were inlined? Are we getting a benefit from extracting the setup of the tests into `let` and `before`?

spec/scope_resolver_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      context 'policy class with not a defined Scope class' do
+        let(:resource_scope) { None }
+
+        before {
+          expect("#{resource_scope}Policy".safe_constantize).to be_present
+          expect("#{resource_scope}Policy::Scope".safe_constantize).to_not be_present
```
dgmstuart: Are these testing our gem code, or are they testing that our test setup is correct? Could something go wrong if we deleted these `before` blocks?

lib/pundit/policy_finder.rb
```diff
@@ -1,53 +1,10 @@
...
-    # @raise [NotDefinedError] if scope could not be determined
-    #
-    def scope_class!
-      raise NotDefinedError, "unable to find policy scope of nil" if resource_scope.nil?
-      scope_class or raise NotDefinedError, "unable to find scope for `#{policy}` policy"
-    end
-  end
-
-  # Finds policy and scope classes for given object.
+  # Finds policy for given object.
```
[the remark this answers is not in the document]
laertispappas: @dgmstuart Fixed

dgmstuart:
@laertispappas Hi - I now work for Varvet, so maybe you could try rebasing and I'll see what I can do about pulling this through

Linuus:
Is this a duplicate of #391?

laertispappas:
@Linuus  Yes it is.

laertispappas:
Duplicate of: https://github.com/varvet/pundit/pull/391

Linuus:
Closing in favor of #391

PR #488, opened 2017-10-11
Use `.all` in generated policy scopes

iangreenleaf:
The current generated policy scopes return the `scope` object directly. This can cause some weird inconsistent behavior, since this means that `policy_scope(MyKlass.scope_or_relation)` will return a relation-like object, while `policy_scope(MyKlass)` will return a class, which is not a relation and does not respond appropriately to things like `my_policy_scope.to_a` or `MyKlass.where(xyz).merge(my_policy_scope)`.

I think, and the examples in the README agree, that for at least everyone using ActiveRecord the best default is to return `scope.all`. This patch updates the generated policies to use that default. I also updated a spec helper that was using the same pattern—not crucial, but it avoids misleading people and I think I improved the specs at the same time.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -87,10 +87,21 @@ module Customer
   end
 end
 
+class CommentScope
+  attr_reader :original_object
+  def initialize(original_object)
+    @original_object = original_object
+  end
+
+  def ==(other)
+    original_object == other.original_object
+  end
+end
+
 class CommentPolicy < Struct.new(:user, :comment)
   class Scope < Struct.new(:user, :scope)
     def resolve
-      scope
+      CommentScope.new(scope)
     end
   end
 end
--- a/lib/generators/pundit/install/templates/application_policy.rb
+++ b/lib/generators/pundit/install/templates/application_policy.rb
@@ -43,7 +43,7 @@ class ApplicationPolicy
     end
 
     def resolve
-      scope
+      scope.all
     end
   end
 end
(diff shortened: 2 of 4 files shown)


timcheadle:
Just ran into this and came up with the same solution. 👍

spec/spec_helper.rb
```diff
@@ -83,10 +83,21 @@ def policy_class
...
+  def ==(other)
+    original_object == other.original_object
+  end
+end
+
 class CommentPolicy < Struct.new(:user, :comment)
   class Scope < Struct.new(:user, :scope)
     def resolve
-      scope
+      CommentScope.new(scope)
```
Linuus: Why did you change this?

iangreenleaf: This spec helper was using the old pattern of returning `scope` directly. I didn't think it made sense to leave that as example code when we're discouraging that pattern's use. I updated it to return a simple mock scope object, so we can test that the policy is correctly wrapping the object it's given.

Linuus: Hmm, I think returning this CommentScope object makes the test more confusing that to just return the `scope`. In the test we just want to know that the `resolve` method returns the expected value, whatever it is.

Can you revert the `CommentScope` code for now and I'll merge the generator fix?

Linuus: Or nah... Let's keep it as it is :) Thanks!

Linuus:
This looks fine to me!

Can you just rebase on master so we get a green Travis build?

iangreenleaf:
Done, thanks for the review!

Linuus:
Thanks!

PR #491, opened 2017-10-30
Add PolicyFinder spec and refactor to avoid rescues

brendanthomas1:
Small refactor of `PolicyFinder` to use ActiveSupport's [`safe_constantize`](https://apidock.com/rails/ActiveSupport/Inflector/safe_constantize) method instead of having to rescue `NameError`.

Also adds a `policy_finder_spec` for unit tests of the `PolicyFinder` class. 

I'm not sure how you all feel about stylistic refactors 🤷‍♂️ . If you're not a fan, I can remove the second commit here. I think the tests add value, though.

### DIFF
--- a/lib/pundit/policy_finder.rb
+++ b/lib/pundit/policy_finder.rb
@@ -23,9 +23,7 @@ module Pundit
     #   scope.resolve #=> <#ActiveRecord::Relation ...>
     #
     def scope
-      policy::Scope if policy
-    rescue NameError
-      nil
+      "#{policy}::Scope".safe_constantize
     end
 
     # @return [nil, Class] policy class with query methods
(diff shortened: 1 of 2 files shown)


Linuus:
@brendanthomas1 Thank you for the contribution.

Can you rebase on master so we get a green build?

lib/pundit/policy_finder.rb
```diff
@@ -23,9 +23,7 @@ def initialize(object)
     #   scope.resolve #=> <#ActiveRecord::Relation ...>
     #
     def scope
-      policy::Scope if policy
-    rescue NameError
-      nil
+      "#{policy}::Scope".safe_constantize
```
Linuus: Could you add an `if policy` here like it was before?

Linuus: Well, I guess it doesn't really matter when we use `.safe_constantize`...

brendanthomas1: yep! `policy` will return either *something* or `nil` and 
```ruby
"#{nil}::Scope".safe_constantize"
=> nil
```

brendanthomas1:
@Linuus rebased off of master

Linuus:
Thanks!

Issue #490, opened 2017-10-23
Different exception handling for NameError


DanielOrtel:
Hey, loving this gem, but there is one issue that I would like to bring up, and that is rescuing NameErrors. For reference here's a method that caused quite a few headaches for me: 
```
def policy
  klass = find
  klass = klass.constantize if klass.is_a?(String)
  klass
rescue NameError
  nil
end
```
The problem with this is that it not only rescues NameError, but it also rescues all child classes of NameError(like NoMethodError) without re-raising the original exception or providing an exception stack. There were a couple of occasions when I called an undefined method from one of the query methods, and kept getting nondescript NotDefinedError-s for, scratching my head at what, where could've gone wrong, until I realized that I forgot to define a method that I called. 

Is there chance that this could be updated, so that it only catches NameErrors and not any of it's children, like, say 

```
rescue NameError => e
  raise NotDefinedError, "msg" if e.class.eql? NameError
  raise
end
```
Or any other solution?

jarosluv:
@DanielOrtel 

Please, see my PR for your issue. Now we need to come up with suitable test cases.

brendanthomas1:
Also hit this in #491 to avoid [rescuing at all](https://github.com/elabs/pundit/pull/491/commits/d6187332d80a8e43730b85bdaf2b600f1c71eb5c) 🤷‍♂️

jarosluv:
@brendanthomas1 

Ah hell! Pundit got ActiveSupport as dependency! I didn't know this before. Thank you opened my eyes :) And now I think we need to remove all active support related code if it possible. Let's discuss here #493.

But your tests, I love it! 👍

PR #492, opened 2017-10-31
[WIP] Fix NameErrors catching in PolicyFinder

jarosluv:
Tried to implement clear handling of catching NoMethodError in accordance with #490 .

Now we need to add suitable test cases.


Linuus:
@jarosluv What's the status here? I think #447 will fix this issue as well?

Issue #178, opened 2014-07-21
Unable to access namespaced policy in view


jrmehle:
Given a TagsController and an Admin::TagsController each with a corresponding policy, how does one access the admin policy from a view? I've got a global nav link I'd like to only expose to admins.

`policy(Tag).index?` references the non-Admin namespace policy

I assumed something like this would work:  `policy(Admin::Tag).index?` However, Admin::Tag is not defined and thus, it throws an error.

Ideally, I'd do something like so: `policy(Tag, namespace: 'admin').index?`

Systho:
I've been using this module to solve my namespace problem. It is not covered by automated tests (but is used on production ) and I just put it here so that it might help other people. Feel free to use it, change it or say it's crap, it has served me well and it might help you, too :)

https://gist.github.com/Systho/3d7632b5aa999cf88d87

hwhelchel:
What about using simple delegators? 

Let's say we are a news media app with a paywall for articles so we want have a policy for article preview and for the full article. 

```ruby
class Api::ArticlePreviewsController < Api::BaseController

  def index
    respond_with :api, article_previews, each_serializer: ArticlePreviewSerializer
  end

  def show
    respond_with :api, article_preview, serializer: ArticlePreviewSerializer
  end

  private

  def article_previews
    @article_previews ||= policy_scope ArticlePreview.new(Article)
  end

  def article_preview
    @article_preview ||= authorize ArticlePreview.new(Article.find(params[:id]))
  end
end
```

```ruby
class ArticlePreview < SimpleDelegator; end
```

Since ArticlePreview is a delegator it will delegate to anything passed in the initializer so that is why we can pass in the class or the instance of the Article. However if you ask an instance of `ArticlePreview` what it's `#class` is it returns `ArticlePreview`

rapind:
My 2 cents are that I'm convinced that authorization should be tied to controllers and not models. To that end, namespacing would be nice.

However if the Pundit philosophy is of a model based authorization, then it's almost definitely a bad idea to add namespacing.

chendrix:
Even if we believe that Pundit should be tied to authorization around models and not controllers, **there is such a thing as namespaced models**.

chendrix:
As a potential work around for those trying to get namespaced policies in views, you can do something like:

If you have a `Foo::BarPolicy` and you're trying to access the `show?` method

```Ruby
policy('Foo::Bar'.to_sym).show?
```

pascalbetz:
[...]

PR #497, opened 2017-11-20
[Update Readme Only] Add namespace policy example

wayne5540:
Thanks for the hard work on this amazing project. It's would be great if we add an example to specify how to override customize policy finder so that we can do a controller based authorization. references: #12, #178

I've posted my solution on [issue12](https://github.com/elabs/pundit/issues/12#issuecomment-345693675), I think it would be great to list on README.md as well since I'm not the only one want to do controller based authorization. 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,8 @@ end
 
 ## Manually specifying policy classes
 
+### From a given class
+
 Sometimes you might want to explicitly declare which policy to use for a given
 class, instead of letting Pundit infer it. This can be done like so:
 
@@ -358,6 +360,38 @@ class, instead of letting Pundit infer it. This can be done like so:
 class Post
   def self.policy_class
     PostablePolicy
+  end
+end
+```
+
+### From the place you call `authorize`
+
+Sometimes you might want to apply namespace for your policy, especially when you
+have namespace for your controller, you can simply override `policy(record)` method
+from you controller.
+
+Here is an example for rails controllers:
+
+```ruby
+# app/controllers/api/v1/base_controller.rb
+class Api::V1::BaseController < ApplicationController
+  include Pundit
+
+  private
+
+  def policy(record)
+    policies[record] ||= "#{controller_path.classify}Policy".constantize.new(pundit_user, record)
+  end
+end
+```
+
+And add your policy
+
+```ruby
+# app/policies/api/v1/auth/offer_policy.rb
+class Api::V1::Auth::OfferPolicy < Api::V1::BasePolicy
+  def index?
+    true
   end
 end
 ```


maxshelley:
I don’t think you need to over-ride the internal method to achieve what you're looking to do. Does passing the namespace in as a symbol when you need to check the policy achieve the same result?

With a `Admin::PhotosController` here's how that would look:

``` ruby
  class Admin::PhotosController < ApplicationController
    def show
      authorize [:admin, @photo], :show?
    end
  end
```

Which would look for `app/policies/admin/photo_policy.rb`.

That would look like:

``` ruby
  class Admin::PhotoPolicy < ApplicationPolicy
    def show?
      # perform auth logic e.g. user.admin?
    end
  end
```

This approach seems to be working for us.

If you didn't want to specify the namespace each time and would rather infer it from the namespace of the controller, you could add something to the application controller that would generate the array of symbols and then pass in the object you're authorizing. I haven't had a chance to test this, but I think it would look like:

``` ruby

  # ...
  
  private

  def policy_with_namespace(object_to_auth)
    controller_path.split('/').map(&:to_sym) << object_to_auth
  end
end
```

_I'm pretty sure that would produce a plural for the name of the policy, but you could always call [.singularize](https://apidock.com/rails/ActiveSupport/Inflector/singularize) on the last item in the array._

You could then have this in a controller:

``` ruby
  def show
    authorize policy_with_namespace(@photo), :show?
    # ...
  end
```

I do agree that having something about namespacing in the docs would be useful, I had to look around to find the implementation we ended up using.

wayne5540:
Thanks @maxshelley, very helpful tip! I didn't know it's able to use namespace to authorize, I think it's a good approach because no method has been overrode. However, override `policy(record)` works great for me as well :)

And yes, my point is there should be some docs about namespace, I had to dig source code to find the implementation, and it took time... 😅

edusantana:
@maxshelley how about this:

if you have an Admin namespace, you should have a AdminController for that namespace.

```ruby
class Admin::AdminController < ApplicationController
end

class Admin::PhotosController < Admin::AdminController # instead of ApplicationController
end
```

Now, we have to know when we at are the admin path. We coulddo that from the controller. The ideia here is to add a dynamic method (`at_admin?`) to the instance with our context

```ruby
  include Pundit

  def pundit_user
    def current_user.at_admin?
      false
    end
  end

end

class Admin::AdminController < ApplicationController
  def pundit_user
    def current_user.at_admin?
      true # every subclass will have this context true
    end
  end
end
```

Now, at the policy files i would be able to call `user.at_admin?` and know my context.

We can even an organization:

```ruby
  include Pundit

  before_action :get_organization

  def pundit_user
    def current_user.at_admin?
      false
    end
    # adding organization to context
    def current_user.current_org
      @org
    end

  end

  private
  def get_organization
    @org = ...
  end
end
```

That's what i'm **trying** to do.

maxshelley:
@edusantana I've read that a couple of times, but I'm not sure what you're trying to achieve with the context you're trying to add. If you need something in the policy, I suggest either passing it as part of the array in `authorize`, or using the additional context solution suggested in the [readme](https://github.com/varvet/pundit#additional-context).

Although I would bear this in mind (from the readme):

> Pundit strongly encourages you to model your application in such a way that the only context you need for authorization is a user object and a domain model that you want to check authorization for. If you find yourself needing more context than that, consider whether you are authorizing the right domain model, maybe another domain model (or a wrapper around multiple domain models) can provide the context you need.

> Pundit does not allow you to pass additional arguments to policies for precisely this reason.

Hope that's useful.

edusantana:
@maxshelley the question was: how to authorize with namespaces? Supose we have two controllers (one without namespace, and the other with `admin`), and one model.

The pundit way is to have only one Policy file, since it only have one model.

When authorizing, how to know if we are at :admin of not? My solution was to use the controller to add a method `at_admin` to users.

maxshelley:
> The pundit way is to have only one Policy file, since it only have one model.

@edusantana I don't think this is the case, I think when it says “the only context you need for authorization is a user object and a domain model that you want to check authorization for” it means “only auth one model at a time”, not “you should only have one policy file”.

Using the `Photo` model from my earlier example, here's how we're working.

In the controllers, we add a namespace to the array that's passed to `authorize`.

``` ruby
class PhotosController < ApplicationController
  def show
    authorize [@photo], :show?
  end
end
```

``` ruby
class Admin::PhotosController < ApplicationController
  def show
    authorize [:admin_policy, @photo], :show?
  end
end
```

_Note: we use `:admin_policy` rather than `:admin` as we had issues with namespaced policies clashing with namespaced models._

Then we end up with two policy classes. Note the different paths:

``` ruby
# app/policies/photo_policy.rb
class PhotoPolicy < ApplicationPolicy
  def show?
    # perform auth logic e.g. user.can_see_photo?
  end
end
```

``` ruby
# app/policies/admin_policy/photo_policy.rb
class Admin::PhotoPolicy < ApplicationPolicy
  def show?
    # perform auth logic e.g. user.admin?
  end
end
```

We're still only auth'ing one model at a time, we change the policy we're pointing towards based on the namespace. This seems to be working well for us so far.

edusantana:
@maxshelley I see.

We will have to add this `[:admin_policy, ...]` to every controller action. If we have many models, and controllers, it will be at many places.

Instead of that, I suggest we could read from the controller's module name:

```ruby 
class PhotosController < ApplicationController
  def show
    authorize @photo, :show?
  end
end
class Admin::PhotosController < ApplicationController
  def show
    authorize @photo, :show?
  end
end
```

Then we end up with two policy classes:

```ruby
# app/policies/photo_policy.rb
class PhotoPolicy < ApplicationPolicy
  def show?
    # perform auth logic e.g. user.can_see_photo?
  end
end
# app/policies/admin/photo_policy.rb
class Admin::PhotoPolicy < ApplicationPolicy
  def show?
    # perform auth logic e.g. user.admin?
  end
end
```

There's no need for this `:admin_policy`, we should read from the module name: `Admin::...`

Do you agree?

maxshelley:
> Instead of that, I suggest we could read from the controller's module name:
> There's no need for this :admin_policy, we should read from the module name: Admin::...
> Do you agree?

@edusantana I did reference this in the last part of my comment here: https://github.com/varvet/pundit/pull/497#issuecomment-348677681

I don't think this needs to be in Pundit itself as you can implement this at your own end. Leaving it out allows us each to decide on our own implementation.

I much prefer the explicit nature of seeing the policy being called, the additional line in each action is well worth the trade-off for us.

Linuus:
I think the documentation about namespaces that I just merged is enough though: 
https://github.com/varvet/pundit#policy-namespacing

PR #501, opened 2018-01-15
spring stop is needed

edusantana:
If we are only runing tests, we don't have a server to restart, but we may need to restart spring as well.


ce07c3:
Hm, this doesn't happen to me.

Linuus:

I think this only happens if you add new folders, right? Otherwise rails/spring should be able to pick it up. I don't think this has to be added in Pundit's readme though.

PR #508, opened 2018-04-16
Pin RuboCop version to 0.51.0

meganemura:
Build is failing due to update of RuboCop now.
I think pinning RuboCop version is friendly to contributors.


meganemura:
0.51.0 is the version that pundit build is succeeded.

Linuus:
Thank you for the PR but this has now been fixed without pinning the version.

PR #466, opened 2017-06-17
Remove wrong statement for what #authorize returns

mackshkatz:

PR #511, opened 2018-04-24
Revert "Remove wrong statement for what #authorize returns"

Linuus:
Reverts varvet/pundit#466

The readme was actually correct for the code in master. A new version of the gem should be released instead. However, I'll need some time to go through all the old issues and PR's before that will happen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,15 @@ def admin_list
 end
 ```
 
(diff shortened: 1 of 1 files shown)


halo:
@Linuus Thank you!
@ce07c3 🎉

PR #514, opened 2018-04-30
Update README.md

seanfcarroll:
Adds documentation around how to apply Scopes to individual actions. Based on [this](https://stackoverflow.com/a/25609167/993592) StackOverflow answer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,38 @@ You can, and are encouraged to, use this method in views:
 <% end %>
 ```
 
+A Scope can be applied to a single record, for example the ```show``` method without duplicating code between the scope and the individual actions. This can be either done in the the controller or within the Pundit policy.
+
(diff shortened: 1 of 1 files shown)


Linuus:
In the index action you should use `pundity_scope(Customer)` or whatever you want to do.
In a show action you should use `@customer = pundity_scope(Customer).find(params[:id])`. The scope should narrow the search down to the records you are allowed to access alread.
Now when you have the record you do the usual `authorize @customer`. In the `show?` method in your policy you should still check that the user is allowed to view the record like

```ruby
def show?
  return false unless user.has_role?(:admin)
  record.organization_id == user.organization_id
end
```

Or whatever you want to check, this is just an example :)

I don't think we need all this documentation around it, however I'll add an example using a scope to find a single record in the scoped section.

PR #517, opened 2018-05-15
Refactor change namespace from Pundit to Rails

yhirano55:
This pull request is re-opened to resolve only refactoring issues (related #473).

## Summary

Changed generator namespace from `rails g pundit:policy something` to `rails g policy something`. Because I think that Policy is custom layer for Rails application same as Decorator ([active_decorator](https://github.com/amatsuda/active_decorator#usage), [draper](https://github.com/drapergem/draper#generators)).

/cc @dgmstuart @Linuus


Linuus:
Hi!

I like having the namespace. This avoids name clashes with other libraries or core features. A lot of gems use namespaces.

Examples:
https://github.com/thoughtbot/scenic
https://github.com/thoughtbot/clearance
https://github.com/plataformatec/devise

So, I think we'll keep the namespace for now.

Thanks!

yhirano55:
Thanks. I understood.

PR #519, opened 2018-05-24
Travis: jruby-9.2.0.0

olleolleolle:
This PR updates the CI matrix to use latest JRuby.

http://jruby.org/2018/05/24/jruby-9-2-0-0.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,6 @@ matrix:
     - rvm: 2.3.5
     - rvm: 2.4.2
     - rvm: 2.5.1
-    - rvm: jruby-9.1.8.0
+    - rvm: jruby-9.2.0.0
       env:
         - JRUBY_OPTS="--debug"


Linuus:
@olleolleolle Thanks! Do we want to run both versions on CI? I have no real experience with jruby so I'm not sure how big the difference is between 9.1.8 and 9.2.0...

olleolleolle:
@Linuus They're wildly different, but be brave, support the latest generally available.

JRuby 9.2.0.0 is Ruby 2.5, for example.

olleolleolle:
@Linuus Being more clear: I think you only need to test on one version of JRuby, the latest generally available. That's the one which has the full attention of the JRuby team.

Linuus:
Thank you for the clarification. However, since jruby 9.2 was only released yesterday I think I'm going to keep 9.1.8 on the CI for a bit before moving to 9.2 only.

FIxed in: https://github.com/varvet/pundit/commit/c4b29bed99f974052069899cace31315bd1f921a

PR #522, opened 2018-06-01
Improve model policy template convenience

jonathanhefner:
With an ApplicationPolicy set up to deny all, the first step to take after generating a model policy is to override ApplicationPolicy methods.  This patch adds stub methods to the model policy template to make that step more convenient.  The `new?` and `edit?` methods are excluded because the default ApplicationPolicy already defers their behavior to `create?` and `update?`, respectively.

### DIFF
--- a/lib/generators/pundit/policy/templates/policy.rb
+++ b/lib/generators/pundit/policy/templates/policy.rb
@@ -1,5 +1,25 @@
 <% module_namespacing do -%>
 class <%= class_name %>Policy < ApplicationPolicy
+  def index?
+    super
+  end
+
+  def create?
+    super
+  end
+
+  def update?
+    super
+  end
+
+  def delete?
+    super
+  end
+
+  def permitted_attributes
+    %i[]
+  end
+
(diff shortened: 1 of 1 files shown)


jonathanhefner:
@Linuus  I'm curious why you say that.  Do most people not configure their ApplicationPolicy to deny all by default?

Linuus:
@jonathanhefner Sure? But what does methods that only call `super` have to do with that? A method that only calls `super` is exactly the same as having no method at all. And I prefer having no method at all in that case :)

jonathanhefner:
@Linuus  The methods are intended to be stubs.  i.e. After generating the model policy, you do change all `super`s to your policy logic (ditto for `permitted_attributes`).  The stubs are provided to reduce typing.

Thank you for answering, by the way.  Your recent work on Pundit is appreciated.

Linuus:
Yes I understand that. I just think they are more noise than help I’m afraid. But I’m really grateful for your contribution still!

Thank you, you’re welcome :)

PR #525, opened 2018-06-07
Allow specification of a NilClassPolicy

bendilley:
This saves having to check whether objects are nil before checking their
policy.

This contradicts the intent of elabs/pundit#251 and could represent a
breaking change for some.

@Linuus a couple of our projects are using the branch I created the original PR from, so I'm opening this rebased as a separate PR.

### DIFF
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -67,8 +67,8 @@ describe Pundit do
       expect(Pundit.policy_scope(user, Article)).to be_nil
     end
 
-    it "returns nil if blank object given" do
-      expect(Pundit.policy_scope(user, nil)).to be_nil
+    it "raises an exception if nil object given" do
+      expect { Pundit.policy_scope(user, nil) }.to raise_error(Pundit::NotDefinedError)
     end
 
     it "raises an error with a invalid policy scope constructor" do
@@ -98,7 +98,7 @@ describe Pundit do
     it "throws an exception if the given policy scope is nil" do
       expect do
         Pundit.policy_scope!(user, nil)
-      end.to raise_error(Pundit::NotDefinedError, "unable to find policy scope of nil")
+      end.to raise_error(Pundit::NotDefinedError, "Cannot scope NilClass")
     end
 
     it "raises an error with a invalid policy scope constructor" do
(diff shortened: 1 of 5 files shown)


spec/pundit_spec.rb
```diff
@@ -409,7 +409,7 @@
     end
 
     it "raises an error when the given record is nil" do
```
Linuus: Do we need this test at all if we allow the nil policy? Isn't it just testing now that it raises if the permission check fails, which is what we do in the test on line 397?

bendilley: I agree that this is now simply testing the NilPolicy, but it could also help prevent a re-implementation of #251 by documenting the intention. I suppose it could read
```ruby
it "denies authorization to destroy a nil object" do
```
or similar.

There's another argument to suggest we shouldn't test the nil case at all, based on the principle that `nil`, like any other object, should be allowed a policy (or lack of one) which functions however the implementer intends.

Linuus: Yeah, I meant we could remove it all together since now we treat `nil` as any other object. But let's keep it for now.

Linuus:
Thanks!

arpitchauhan:
Good work @bendilley

connorshea:
I'm not sure if I should open an issue for this instead of commenting here, tell me if I should move this to a separate issue :)

So I have some controller actions that are non-standard, e.g. 

```erb
<%# view %>
<% if policy(current_user).add_release_to_library? %>
  Button goes here
<% end %>
```

```rb
# policy
class UserPolicy < ApplicationPolicy
  attr_reader :current_user, :user

  def initialize(current_user, user)
    @current_user = current_user
    @user = user
  end

  def add_release_to_library?
    current_user && user == current_user
  end
end
```

and then a NilClassPolicy:

```rb
class NilClassPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      raise Pundit::NotDefinedError, "Cannot scope NilClass"
    end
  end

  def index?
    false
  end

  def show?
    false
  end

  def create?
    false
  end

  def update?
    false
  end

  def destroy?
    false
  end
end
```

The problem I'm having is that users which aren't logged in have `current_user` set to `nil`, which causes `policy(current_user).add_release_to_library?` to fallback to the NilClassPolicy, which doesn't have a method called `add_release_to_library?`. It gives me an error like this: `NoMethodError: undefined method `add_release_to_library?' for #<NilClassPolicy:0x00007ff0c2c8a478 @user=nil, @record=nil>`

I know I could just add all these non-standard methods to NilClassPolicy and call it a day, but that feels like a really dumb way of doing things. Similarly, I could include a check for `current_user` in every view, but that also feels really dumb.

Is there a way to have Pundit catch all other methods for NilClassPolicy and return false?

PR #529, opened 2018-06-17
Pass last element in array to policy and scope

Linuus:
Closes #391 #351

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -80,7 +80,7 @@ module Pundit
     # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
     def policy_scope(user, scope)
       policy_scope = PolicyFinder.new(scope).scope
-      policy_scope.new(user, scope).resolve if policy_scope
+      policy_scope.new(user, pundit_model(scope)).resolve if policy_scope
     rescue ArgumentError
       raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
     end
@@ -95,7 +95,7 @@ module Pundit
     # @return [Scope{#resolve}] instance of scope class which can resolve to a scope
     def policy_scope!(user, scope)
       policy_scope = PolicyFinder.new(scope).scope!
-      policy_scope.new(user, scope).resolve
+      policy_scope.new(user, pundit_model(scope)).resolve
     rescue ArgumentError
       raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
     end
@@ -124,7 +124,7 @@ module Pundit
     # @return [Object] instance of policy class with query methods
     def policy!(user, record)
       policy = PolicyFinder.new(record).policy!
-      policy.new(user, pundit_model(record)) if policy
+      policy.new(user, pundit_model(record))
     rescue ArgumentError
       raise InvalidConstructorError, "Invalid #<#{policy}> constructor is called"
     end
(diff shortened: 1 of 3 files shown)


ramaboo:
Thanks for doing this. Sorry I totally fell off the map. I moved companies and we currently use cancancan (which causes me daily grief). Glad to see pundit progressing!

lib/pundit.rb
```diff
@@ -124,10 +124,16 @@ def policy(user, record)
     # @return [Object] instance of policy class with query methods
     def policy!(user, record)
       policy = PolicyFinder.new(record).policy!
-      policy.new(user, record)
+      policy.new(user, pundit_model(record))
```
gagalago: Sorry to come soo late, but with this implementation, we are not able anymore to have the namespace of the policy. 

eg: `@parent` in case of `authorize [@parent, @childrens]`

what do you think to splat the record instead:
```
if record.is_a?(Array)
  policy.new(user, *record)
else
  policy.new(user, record)
end
```

Linuus: Do you mean that the issue is that you can’t pass multiple things to your policy? This is not the intention of passing an array, it’s merely for namespacing. If you need multiple objects such as parent/child objects, wrap them in another object.

gagalago: strange that I have to do somethings like that:
```
authorize [@parent, MyCustomClass.new(@parent, @childrens)]
class MyCustomClass
  attr_accessor :childrens
  delegates :table_name, to: :childrens
  def initialize(parent, childrens)
    self.childrens = childrens
  end
end
class Parent::ChildrenPolicy < ApplicationPolicy
   ...
end
```

I understand that when you provide symbol to create a namespace, that kind of value are not interesting but when you provide object, that can be intersting to be able to retrieve them. It's why I propose to give to the policy all resources and to edit the ApplicationPolicy accordingly
```
class ApplicationPolicy
  def initialize(user, *namespace, subject)
    @user = user
    @namespace = namespace
    @subject = subject
  end
end
```
So that if I want, I have access to `parent` in form of an array and I didn't change the signature for `subject`.

Linuus: But you are using one feature (namespacing) for something that it's not intended for (passing multiple objects for authorization).

1. What if I want multiple objects in my policy. If I have to pass an array, it means I also _have_ to namespace my policy which is weird...
2. The API is weird if using the namespaces feature for passing multiple objects (you suggested solution).
Example:
`authorize [:admin, :publishable, @parent, :whatever, @article]`

In the policy, fetching `@parent` would be awkward since it's mixed in with the namespace stuff. `parent = namespace[2]` or something... This is not an API I want.

gagalago: Depending on the usage, it can be more convenient to pass multiple objects to policy through namespace, sometimes through wrapping them in another object.

what I found weird is that some arguments that I pass to the policy cannot be retrieved inside it.
I use namespace in case of nested routes to check that the nested resource can be owned by the parent resource.

Example from my codebase used for route `/projects/:project_id/nominees/:nominee_id`:
```ruby
class Project::NomineePolicy < NamespacedPolicy
  def update?
    user.can?("admin") || Projects::ProjectPolicy.new(user, [namespace]).own_project?
  end
  class Scope < Scope
    def resolve
      if user.can?("admin") || Projects::ProjectPolicy.new(user, [namespace]).own_project?
        scope.where(project_id: namespace.id)
      else
        scope.none
      end
    end
  end
end
```
It allows me to check that the nominee is still right after edition depending on the namespace or to filter the scope depending on the namespace. As I already wrap these objects into an array I didn't understand why I need also to wrap these objects into another object.

For your example, some specific policies can override initialize to expand the `*namespace` into `admin, publishable, parent, whatever` but I understand that in that specific case, this is pretty ugly.

That's why a propose a way to have a clean way to standard usage (last element) and a less convenient way for the namespace. That show to the developer what is the best way to do it but I didn't prevent the usage of the namespace.

PR #538, opened 2018-07-26
feat(PolicyScope): Add ability to override the policy scope class

### LINKED ISSUE #537: Pundit.policy_scope class method should support new policy_scope_class option
We noticed that the newest release includes this `policy_scope_class` option (which is nice!) for the `policy_scope` method that is included as an instance method on controllers [here](https://github.com/varvet/pundit/blob/v2.0.0/lib/pundit.rb#L236), but it was not added as an option on the class method [here](https://github.com/varvet/pundit/blob/v2.0.0/lib/pundit.rb#L82).

Since we use a mixture of both the instance and class method, depending on the context in which the policy is used, we would like to be consistent in our usage of policy_scope_class. Do you think it makes sense to support this feature in both of these code paths?

kyleholzinger:
There was a new feature added to the instance method `policy_scope`
that allowed the policy scope class to be manually declared via the
`policy_scope_class` argument. This adds that same functionality to
the class method.

Closes #537

### DIFF
--- a/README.md
+++ b/README.md
@@ -545,6 +545,13 @@ Pundit.policy_scope(user, Post)
 The bang methods will raise an exception if the policy does not exist, whereas
 those without the bang will return nil.
 
+Similar to the `policy_scope` instance method, you can also override the policy scope class:
+
+``` ruby
+Pundit.policy_scope!(user, Post, policy_scope_class: PublicationPolicy::Scope)
+Pundit.policy_scope(user, Post, policy_scope_class: PublicationPolicy::Scope)
+```
+
 ## Customize Pundit user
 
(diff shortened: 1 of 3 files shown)


kyleholzinger:
@Linuus I noticed #537 and thought I'd open a PR to realize the implementation. Thoughts? I think it'd be pretty nice to always be able to fetch a policy scope the same way regardless of if you're using the instance method or the class method

sekrett:
You should update Readme in the section "Policy Namespacing" like I did for `authorize` method in https://github.com/varvet/pundit/pull/540.

Linuus:
Closing. See #537 for reasoning.

PR #540, opened 2018-08-03
Update Readme to make possible use of `policy_class` in admin namespace

sekrett:
A little correction. In 2.0 I had to add this to be able to specify `policy_class` in my admin controllers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,8 +582,8 @@ class AdminController < ApplicationController
     super([:admin, scope])
   end
 
-  def authorize(record, query = nil)
-    super([:admin, record], query)
+  def authorize(record, query = nil, options = {})
+    super([:admin, record], query, options)
   end
 end
 


sekrett:
I discovered one important thing. It is good to override the `policy` method so I can use it in views and there is no need to override the `authorize` method, since it relies on `policy` method. If I override both I will get double namespacing. So the controller in the example should be like this:

```
  def policy_scope(scope)
    super([:admin, scope])
  end

  def policy(record)
    super([:admin, record])
  end
end
```

If you agree I will rewrite my PR.

README.md
```diff
@@ -582,8 +582,8 @@ class AdminController < ApplicationController
     super([:admin, scope])
   end
 
-  def authorize(record, query = nil)
-    super([:admin, record], query)
+  def authorize(record, query = nil, options = {})
```
dgmstuart: This isn't quite right: looking at the source this seems to be a specific keyword argument, rather than a generic list of options.

dgmstuart:
Hi @sekrett - I'm sorry this has gone so long unanswered. I don't entirely understand the suggestion, but it looks like https://github.com/varvet/pundit/issues/537 might resolve it?

I'll close this for now, but please re-open if you think there's something worth addressing - thanks.

PR #541, opened 2018-08-04
Policy cache will depend on user also

sekrett:
I have two namespaces with different `pundit_user`.

I noticed that when I do something in admin namespace, then I visit the site (as guest) and come back to admin, sometimes randomly (very hard to reproduce) I get `nil` instead of `user`, which is defined, otherwise devise would redirect me to login page. Much worse it would be if I was logged as admin and as user at the same time, instead of `nil` I would get authorizations against a wrong user!

Error goes away if I restart rails server or change some code to trigger a code reload (in development). I added a `Rails.logger.debug` to my base policy initializers and noticed that policy initializer is not called but I see a SQL query loading the user before, so it is a cache problem. The record may be the same but user is different.

So I added the user as second key, now it works correctly.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -244,7 +244,7 @@ protected
   # @param record [Object] the object we're retrieving the policy for
   # @return [Object, nil] instance of policy class with query methods
   def policy(record)
-    policies[record] ||= Pundit.policy!(pundit_user, record)
+    policies[[pundit_user, record]] ||= Pundit.policy!(pundit_user, record)
   end
 
   # Retrieves a set of permitted attributes from the policy by instantiating
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -427,9 +427,9 @@ describe Pundit do
     end
 
     it "caches the policy" do
-      expect(controller.policies[post]).to be_nil
+      expect(controller.policies[[user, post]]).to be_nil
       controller.authorize(post)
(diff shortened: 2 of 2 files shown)


sekrett:
No, this did not help, but policy initializer is called multiple times now, and the wrong one. I am not sure if it was a good idea to modify the cache keys.

Linuus:
Can you open an issue for this? Also, it would be great if you could create a minimal app where this can be reproduced (even if it’s random).

sekrett:
I have found an error in my app and easy steps to reproduce. I had `ApplicationPolicy` and `Admin::ApplicationPolicy`. If I start rails, then go to site, then to admin, `ApplicationPolicy` is loaded instead of `Admin::ApplicationPolicy`, because class name is the same. After I renamed second class, the bug is gone. So it is reproducible but not related to cache.

It would be hard to invent a mini app to show a possible bug, but you can imagine a situation when `pundit_user` is a method performing some complex logic before returning actual user. I don't know how a user can change in the middle of one request, but it is technically possible. A cache should depend on every parameter to be correct, but you can leave the code as it is since it worked for years.

Linuus:
Yes, this is not related to pundit. I’ve had the same issues with controllers. You can use `require_dependency` in Rails to solve this issue.

Yeah, the cache thing may have some edge cases, however, it should not affect two different requests. 

I’ll close this for now.

PR #550, opened 2018-09-18
Improve exception for #policy_scope and #policy_scope!

### LINKED ISSUE #548: Better Exception on #policy_scope
Hey.

We use `pundit` for a project and we've got an Exception which was confusing.

So we have a Policy with a Scope. And in the scope we did something wrong on a query. The query thrown an exception, and it was an ArgumentError too. But pundit catch it and returned an other exception who wasn't really linked to.

I guess the [`#policy_scope`](https://github.com/varvet/pundit/blob/f6bf57b1b69eb6c859b1a0adb9ba2922a02e4461/lib/pundit.rb#L85) method could be:

```ruby
    def policy_scope(user, scope)
      policy_scope = PolicyFinder.new(scope).scope
    rescue ArgumentError
      raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
    else
      policy_scope.new(user, pundit_model(scope)).resolve if policy_scope
    end
```

or 

```ruby
    def policy_scope(user, scope)
      begin
        policy_scope = PolicyFinder.new(scope).scope
      rescue ArgumentError
        raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
      end

      policy_scope.new(user, pundit_model(scope)).resolve if policy_scope
    end
```

nicolas-brousse:
This changes avoid to raise an `InvalidConstructorError` if we catch an
`ArgumentError` exception from `#resolve` method of Policy Scope.

This PR closes #548.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -80,10 +80,16 @@ module Pundit
     # @raise [InvalidConstructorError] if the policy constructor called incorrectly
     # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
     def policy_scope(user, scope)
-      policy_scope = PolicyFinder.new(scope).scope
-      policy_scope.new(user, pundit_model(scope)).resolve if policy_scope
-    rescue ArgumentError
-      raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
+      policy_scope_class = PolicyFinder.new(scope).scope
+      return unless policy_scope_class
+
+      begin
+        policy_scope = policy_scope_class.new(user, pundit_model(scope))
+      rescue ArgumentError
+        raise InvalidConstructorError, "Invalid #<#{policy_scope_class}> constructor is called"
+      end
+
+      policy_scope.resolve
     end
 
     # Retrieves the policy scope for the given record.
@@ -95,10 +101,16 @@ module Pundit
     # @raise [InvalidConstructorError] if the policy constructor called incorrectly
     # @return [Scope{#resolve}] instance of scope class which can resolve to a scope
     def policy_scope!(user, scope)
-      policy_scope = PolicyFinder.new(scope).scope!
-      policy_scope.new(user, pundit_model(scope)).resolve
-    rescue ArgumentError
-      raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
+      policy_scope_class = PolicyFinder.new(scope).scope!
+      return unless policy_scope_class
+
+      begin
+        policy_scope = policy_scope_class.new(user, pundit_model(scope))
+      rescue ArgumentError
+        raise InvalidConstructorError, "Invalid #<#{policy_scope_class}> constructor is called"
+      end
+
+      policy_scope.resolve
     end
 
     # Retrieves the policy for the given record.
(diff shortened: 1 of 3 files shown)


nicolas-brousse:
Let me know if you want some changes on specs. I could do some updates but not sure what's the best way to do it.

lib/pundit.rb
```diff
@@ -80,10 +80,13 @@ def authorize(user, record, query, policy_class: nil)
...
     # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
     def policy_scope(user, scope)
-      policy_scope = PolicyFinder.new(scope).scope
+      begin
+        policy_scope = PolicyFinder.new(scope).scope
+      rescue ArgumentError
+        raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
+      end
+
       policy_scope.new(user, pundit_model(scope)).resolve if policy_scope
```
Linuus: As we discussed in the PR #548 the call we want to rescue errors from is `policy_scope.new(user, pundit_model(scope))`. So you have to move that into the rescue block and only run the `resolve` call outside of it.

lib/pundit.rb
```diff
@@ -95,10 +98,13 @@ def policy_scope(user, scope)
...
-      policy_scope.new(user, pundit_model(scope)).resolve
-    rescue ArgumentError
-      raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
+      begin
+        policy_scope = PolicyFinder.new(scope).scope!
+      rescue ArgumentError
+        raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
+      end
+
+      policy_scope.new(user, pundit_model(scope)).resolve if policy_scope
```
Linuus: Same as the above comment.

lib/pundit.rb
nicolas-brousse: I pushed some code with what I said. Let me know if it's good for you.

lib/pundit.rb
```diff
@@ -80,10 +80,15 @@ def authorize(user, record, query, policy_class: nil)
...
-      policy_scope = PolicyFinder.new(scope).scope
-      policy_scope.new(user, pundit_model(scope)).resolve if policy_scope
-    rescue ArgumentError
-      raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
+      policy_scope_class = PolicyFinder.new(scope).scope
+
+      begin
+        policy_scope = policy_scope_class.new(user, pundit_model(scope))
+      rescue ArgumentError
+        raise InvalidConstructorError, "Invalid #<#{policy_scope}> constructor is called"
```
Linuus: The `policy_scope` variable in the error message should be `policy_scope_class`. This is why the tests are failing.

nicolas-brousse:
Thanks for the review. Sorry I didn't look enough the tests and I didn't saw the failure.

I update the specs and resolve some mistakes I made in my code. I didn't care but I keep a condition that was at the end, but it needs to be done at first now (`return unless policy_scope_class`).

If you don't like this kind of syntax I can do an update. And finally for specs, I created a Thread object and policy.

Let me know if my last changes are good for you.

nicolas-brousse:
There is error during spec execution for ruby versions < 2.3 and jruby. `ArgumentError` messages change. Do you think we could just check the exception without its message?

Linuus:
Hm, that's annoying :) Let's remove the check on the message for `ArgumentError` then as you proposed.

Can you also please squash all commits in this PR and force push?

nicolas-brousse:
I'll do the changes for `ArgumentError` so.

Ok for squash all my commits together.

Does README or CHAGELOG needs to be updated too? Or all's good?

Linuus:
Thanks!

No, I can update the changelog later after merge.

nicolas-brousse:
All's green 🎉

PR #553, opened 2018-10-11
Add Suffix Configuration

geshafer:
Pundit uses a `SUFFIX` constant for locating Policy classes inferred
from objects passed to the authorization helpers. It is very helpful to
override this suffix in cases where the `Policy` keyword is not
preferred.

As an example, in an application where insurance policies are managed,
introducing the `Policy` keyword to the domain language can cause logic
to get confusing.

To allow applications to override this suffix without removing and
reassigning the constant, this commit introduces a configuration class
to the gem that allows developers to define an configuration block in an
initializer to override this value.

This could be expanded upon later by either adding things to the
configuration or dynamically defining the `policy` and `policy!` helpers
to instead use the configured suffix.

Appropriately this commit also updates the documentation to describe
configuring this suffix which remains `Policy` by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,6 +599,33 @@ class Admin::PostController < AdminController
 end
 ```
 
+## Configuring Custom Policy Names
+
+In very rare occassions the `Policy` suffix may not work well in your
+application domain. To resolve this issue, Pundit allows you to
+configure a different suffix for your application policies so that the
+policy lookup well infer policy names using that suffix rather than
+`Policy`.
+
+```ruby
+Pundit.configure do |config|
+  config.suffix = "Authorization"
+end
+
+class PolicyAuthorization < ApplicationAuthorization
+  def show?
+    user.admin?
+  end
+end
+
+class Admin::PolicyController < AdminController
+  def show
+    policy = Policy.find(params[:id])
+    authorize(policy)
+  end
+end
+```
+
 ## Additional context
 
 Pundit strongly encourages you to model your application in such a way that the
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "pundit/version"
+require "pundit/configuration"
 require "pundit/policy_finder"
 require "active_support/concern"
 require "active_support/core_ext/string/inflections"
@@ -10,8 +11,6 @@ require "active_support/dependencies/autoload"
 
 # @api public
 module Pundit
-  SUFFIX = "Policy".freeze
-
   # @api private
   module Generators; end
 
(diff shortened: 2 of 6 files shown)


Linuus:
Thank you for the contribution @geshafer!

I could see this being useful in some edge cases. However, I don't think I want to merge this into Pundit though, for a couple of reasons.

1. It makes the code a lot more complex to only solve quite a small edge case.
2. Documentation would be harder since we probably cannot just say "policy" everywhere without people being confused when they (or someone on their team) has renamed it to something else.
3. It would be weird to have the helpers and such be called something with `policy` so we would need to generate them dynamically rendering the code even more complex and harder to reason about generating more bugs.

All in all, I'm 👎 on this I'm afraid.

Linuus:
1. Yes, but: Lines of code added is not a good measurement of added complexity. Also, I wouldn't want to _only_ do this but also change all occurrences of `policy`.
2. I'm afraid this will cause confusion.
3. That would be odd in my opinion.

geshafer:
@Linuus I understand your concerns; fortunately, we had a work around but I thought it might be better to make it official to make upgrading easier:

```rb
# config/initializers/pundit.rb
Pundit.send(:remove_const, :SUFFIX)
Pundit::SUFFIX = "Authorization".freeze
```

```rb
# app/helpers/application_helper.rb
def authorization(resource)
  policy(resource)
end
```

We haven't had any issues with not understanding the Authorization is a Policy when working directly with Pundit. I personally think that continuing to use `policy` in the internal API makes sense while allowing the external API to use whatever makes sense. Thanks for looking anyway 😄

ce07c3:
1. I don't understand that. Why does this need to cascade? And the complexity here is low, or what's your measure?
2.+3. If you add the config, you know what you did. It just feels arbitrary to reject this PR.

Linuus:
1. Because I think it’s a bad choice to _not_ change add the other stuff as well. Hence, this PR would never be merged without those changes as well.
2. Scenario: You know nothing about Pundit. You join a team where they have changed this option to `Foobar`. You want to read about these `Foobars` that Pundit supports. This would be confusing...

This change is the same as if Rails would add a config to change controllers to be called something else, like `Managers`, only because in some, very, rare cases someone might have something called `Controller` in their app.

ce07c3:
1. I think it’s totally fine to do that change only. See 2. 
2. You don’t read the manual and check the config before you change stuff? 😇

Also, let’s challenge the last point: You can call your controllers XyzEndpoint if you like and configure Rails to use that: https://guides.rubyonrails.org/routing.html#specifying-a-controller-to-use. 

Rails will not change any method names related to actions and params and so on. So no, I don’t believe that a changed suffix necessitates a renaming of helpers. 

I can see your point about consistency, I just feel it becomes almost an aesthetic virtue here rather than making Pundit easy to use. 😊

Linuus:
1. I have noted your opinion, let's agree to disagree.
2. I don't think it matters what _I_ usually do or do not do. I _do_ think the number of issues and amount of confusion _will_ become higher with a very, very small benefit (it's even noted in this PR that it's only useful in _very rare_ occasions.

We are getting sidetracked here, but no, you cannot rename the controller suffix for Controllers in Rails AFAIK.

The setting you are referring to would look like this, right?

```ruby
resources :foobar, controller: "xyz_endpoint"
```

This will make Rails look for an `XyzEndpointController`.

Issue #542, opened 2018-08-08
"Error" namespace conflicts after including Pundit


mattruzicka:
I updated pundit from 0.3 to 2.0 after ran into the following issue.

In my controllers, the Error constant gets evaluated as Pundit::Error after I include Pundit. This is causing an issue with my own custom errors which are namespaced under my own Error module.

It looks like this particular issue that I'm running into was introduced by this commit https://github.com/varvet/pundit/commit/a21ad899a70374797df1fd905f3966d8f77a6442

Here's an example of what I'm seeing within the context of irb

```
irb(main):001:0> require 'pundit'
=> true
irb(main):002:0> include Pundit
=> Object
irb(main):003:0> Error
=> Pundit::Error
```

I can get around this by doing ::Error::MyCustomError, but I'm reporting this because 1) I don't want to 2) this doesn't seem like the intended behavior.

Moreover, all the error classes defined my pundit have this behavior after including Pundit
```
irb(main):004:0> NotDefinedError
=> Pundit::NotDefinedError
```
This behavior seems like it could lead to a lot of unwanted conflicts. The documentation always references errors by their full name, for example, Pundit::NotAuthorizedError, not NotAuthorizedError and this does not seem to be documented behavior. I would be able to search for a solution that does not make constants like NotDefinedError, NotAuthorizedError, and Error available to classes that include Pundit without namespacing them with Pundit:: if this is desired.

Linuus:
Isn’t this just how Ruby does its constant lookups (which is weird sometimes)?

Linuus:
What do you think about: https://github.com/varvet/pundit/pull/590 ?

dgmstuart:
TODO:

- [ ] release a patch version which uses the compact style for `Pundit::Error` (#590)
- [ ] release a major version which uses a compact style for all error types.

Linuus:
I don’t think we should use compact style for all errors. Instead, we should restructure the code to make this problem go away in a major update.

dgmstuart:
@Linuus OK cool - I'll update the todo. I don't quite understand what restructuring needs to be done, but we can cross that bridge when we come to it.

Burgestrand:
Closed this by merging #621

PR #555, opened 2018-10-20
Change namespace of Pundit::Error -> Pundit::PunditError

kjleitz:
Fixes https://github.com/varvet/pundit/issues/542

Issue introduced by commit https://github.com/varvet/pundit/commit/a21ad899a70374797df1fd905f3966d8f77a6442

**Description:**

Pundit's error classes are namespaced within the `Pundit` module, which is intended to be `include`d into a class for policy functionality. Because of the way Ruby does lookups, that means that any class where you `include Pundit` has the following behavior:

```ruby
class Foo < ActiveRecord::Base
  include Pundit

  def self.bar
    raise Error
  end
end
```

```bash
2.3.3 :001 > Foo.bar
Pundit::Error: Pundit::Error
	from (irb):1
```

The same goes for `NotAuthorizedError`, `AuthorizationNotPerformedError`, etc., but they are of less concern than the common constant `Error`.

**Solution:**

Renamed the `Error` class in the `Pundit` module to `PunditError`.

**Why:**

This doesn't fix the fact that Pundit's errors are still included along with the rest of the module at the same scope as the inclusion. However, the only really nasty one (given that there are just a handful) is `Pundit::Error`, since `Error` is _fairly likely_ to have collisions considering how common the name is.

It doesn't make sense to change the namespacing of the errors (e.g., `Pundit::NotAuthorizedError -> Pundit::PunditErrors::NotAuthorizedError` or similar) because they are part of the public API and would be a breaking change. `Pundit::Error`, on the other hand, is *not* documented as part of the API, and in fact is marked as *private*.

Considering it's private and is only used as a base class from which the others inherit, it is not a breaking change to rename `Pundit::Error` to `Pundit::PunditError`.

**Notes:**

From CONTRIBUTING.md:

> **Add tests!** Your patch won't be accepted if it doesn't have tests.

I haven't included any tests as this is a simple rename, is only used in one file, changes no logic, and tests already cover the relevant child classes (and I've confirmed that all tests still pass)—hope that's acceptable.

> **Document any change in behaviour.** Make sure the README and any other relevant documentation are kept up-to-date.

This error class has always been marked private, and as such is not mentioned in the README nor in the API documentation, so fortunately there is no documentation to update.

**Anyway:**

Thanks for maintaining this gem. It's a pleasure to use 😃

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -16,10 +16,10 @@ module Pundit
   module Generators; end
 
   # @api private
-  class Error < StandardError; end
+  class PunditError < StandardError; end
 
   # Error that will be raised when authorization has failed
-  class NotAuthorizedError < Error
+  class NotAuthorizedError < PunditError
     attr_reader :query, :record, :policy
 
     def initialize(options = {})
@@ -38,18 +38,18 @@ module Pundit
   end
 
   # Error that will be raised if a policy or policy scope constructor is not called correctly.
-  class InvalidConstructorError < Error; end
+  class InvalidConstructorError < PunditError; end
 
   # Error that will be raised if a controller action has not called the
   # `authorize` or `skip_authorization` methods.
-  class AuthorizationNotPerformedError < Error; end
+  class AuthorizationNotPerformedError < PunditError; end
 
   # Error that will be raised if a controller action has not called the
   # `policy_scope` or `skip_policy_scope` methods.
   class PolicyScopingNotPerformedError < AuthorizationNotPerformedError; end
 
   # Error that will be raised if a policy or policy scope is not defined.
-  class NotDefinedError < Error; end
+  class NotDefinedError < PunditError; end
 
   extend ActiveSupport::Concern
 


bjeanes:
+1, though this is likely a breaking change for some people.

ce07c3:
Pundit is supposed to be included in controllers?

kjleitz:
@ce07c3 just a mistake in my example, I'll fix that (and make it clearer). Doesn't change the issue, though!

kjleitz:
Maybe this will make the problem more obvious:

```
2.3.3 :001 > include Pundit
 => Object
2.3.3 :002 > class Error < StandardError; end
 => nil
2.3.3 :003 > raise Error
Pundit::Error: Pundit::Error
	from (irb):3
2.3.3 :004 > raise ::Error
Pundit::Error: Pundit::Error
	from (irb):4
2.3.3 :005 > raise Object::Error
Pundit::Error: Pundit::Error
	from (irb):5
2.3.3 :006 >
```

It doesn't matter if you're in a controller or not, the inclusion simply pollutes the namespace with the generic name `Error`, such that if you include `Pundit` in `MyWhatever`, then `MyWhatever::Error` is now `Pundit::Error`.

Linuus:
What do you think about https://github.com/varvet/pundit/pull/590?

kjleitz:
@Linuus much better than my approach! Closing this in favor of #590 :+1:

Issue #321, opened 2015-10-09
Used ApplicationPolicy if no <MODEL>Policy is present as shared policy, so I do not have to define a Policy file for all my models


fddayan:
The idea is that: 
if you have not define a <MODEL>Policy, it will load `AppliactionPolicy`, that will have the shared policy. This is to avoid `Policy not found` Exception. 

I'm introducing `Pundit` to a new project with a lot of models and I have to created them all manually. Also most of them share the same policy.

ce07c3:
@fddayan Just create a base policy and let all the others inherit from it. Having an implicit default policy is paramount to insecurity and insane defaults. ;)

jnicklas:
Agreed. It also makes debugging much harder.

PR #560, opened 2018-11-02
add class configuration for a fallback_policy

ericsullivan:
I saw this related issue: https://github.com/varvet/pundit/issues/321 and a [SO solution](https://stackoverflow.com/questions/27522622/how-to-use-a-default-app-policies-application-policy-rb) that suggests rescuing NotDefinedError.

Our application has a lot of roles. It means all policies would have a case statement on the role, and each new role would require editing all the policies. For similar roles, we couldn't just copy a folder and tweak a few files, we'd have to open all the polices and say `old_role || similar_role`.

A solution I'm trying is something like this:
```
def authorize(record, query = nil, options = {})
  role = role(pundit_user)

  super([:users, role.to_sym, record].flatten, query, options)
end
```

So we'll defined a policy per model, per role. However, that's also a lot of files. When the policy will just return false, I'd rather not create the policy file. I could rescue the NotDefinedError, but this isn't really an exception. I was thinking a fallback policy like this would work:

```
class Users::FallbackPolicy
  attr_reader :user, :record

  def initialize(user, record)
    @user = user
    @record = record
  end

  def method_missing(method, *args, &block)
    if method[-1] == '?'
      false
    else
      super
    end
  end

  class Scope
    attr_reader :user, :scope

    def initialize(user, scope)
      raise "Not Applicable"
    end
  end
end
```

Is this use case compelling enough to add support to pundit and if so, is this an acceptable implementation?

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -54,6 +54,14 @@ module Pundit
   extend ActiveSupport::Concern
 
   class << self
+    def fallback_policy
+      @fallback_policy
+    end
+
+    def fallback_policy=(policy)
+      @fallback_policy = policy
+    end
+    
     # Retrieves the policy for the given record, initializing it with the
     # record and user and finally throwing an error if the user is not
     # authorized to perform the given action.
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -61,7 +61,7 @@ module Pundit
     def fallback_policy=(policy)
       @fallback_policy = policy
     end
-    
+
     # Retrieves the policy for the given record, initializing it with the
     # record and user and finally throwing an error if the user is not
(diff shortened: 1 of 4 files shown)


Linuus:
It's a bit hard to understand how you've structured your policies at the moment... You say that you have one policy per model and role? That sounds...odd :)

Can't you just have every policy inherit from an `ApplicationPolicy` and define some good defaults for different actions in there? You could even have a separate object which does the role mapping/grouping.

To me this sounds more like an issue of a failed OO architecture which can be solved without adding more complexity to Pundit, but I could be missing something of course :)

Perhaps you can show a more complete example?

Linuus:
Closing due to inactivity.

PR #562, opened 2018-11-08
Fix unvalid ruby code

btrd:


btrd:
My bad, it's valid ruby code, I had an issue but it was a little bit different.

Issue #554, opened 2018-10-12
Unclear what the breaking changes are between 1.0 and 2.0


Gargron:
I'm evaluating whether an upgrade from 1.1.0 to 2.0.0 is safe, but the 2.0.0 changelog does not include any information on which change is breaking. To understand it properly I would then have to read through the entire pundit source code or take a risk. Could you please clarify why you bumped to 2.0.0 and not 1.2.0?

Linuus:
Sure.

These changes could be seen as breaking changes:

- Only pass last element of "namespace array" to policy and scope. (#529)
- Return passed object from #authorize method to make chaining possible. (#385)

And perhaps
- Raise InvalidConstructorError if a policy or policy scope with an invalid constructor is called. (#462)

Is that information enough or do you need me to elaborate on some specific change?

Gargron:
Thank you.

PR #563, opened 2018-11-16
Highlight breaking changes

harigopal:
It's helpful to point out what the breaking updates are when there's a major version change. The list of breaking changes were taken from the response to #554.


Linuus:
Thanks!

Issue #523, opened 2018-06-02
Module `Pundit` is too big


sekrett:
I got a real problem when started to use Pundit. I have my own method `permitted_attributes` which of course conflicts with Pundit same method. I think the included Pundit module should be as small as possible, it should contain only public methods, the rest may go to other classes to avoid polluting `ApplicationController`. I solved the problem by renaming my method, so this is not urgent. Anyway it would be great to refactor.

dmitry:
At least pundit should expose protected method `pundit_permitted_attributes` instead of `permitted_attributes` for the consistency.

gogiel:
I see two separate issues when it comes to Pundit pollution.

## Issue 1: Including top level module

Including top level `Pundit` module has huge side effects. It not only includes methods defined in `Pundit` module, but also adds all internal classes and modules to the scope. And not only the modules that are defined in `pundit.rb` but literally everything! In your controller you can access constants like `VERSION`, SUFFIX` or even `RSpec` from `lib/pundit/rspec.rb`.

Solution for that is extracting only the public interface to a module inside `Pundit` global module, e.g. `Pundit::ControllerExtension`. In order to prevent compatibility change `Pundit` global module could still work the old way, by including new module (`ControllerExtension`) in `self.included()` with a deprecation message.

Fixing it should be rather simple and backward compatible. I'll give it a try and prepare a PR.

## Issue 2: Unexpected methods pollution

Pundit "pollutes" both class and instance scopes with some methods that are not prefix with `pundit_`. As it's been noted in this Issue it might force some users to rename their own methods and lead to unexpected behaviours, especially for junior devs who may be completely unaware of that.

I'd suggest at least prepending all the methods with `pundit_` prefix to make that clear.

That would require introducing breaking changes to the interface. Before starting working on it I would love to have a clear consensus from the original devs of this gem about this breaking change.

### Methods pollution in general

Ideally Pundit could avoid method pollution when included in general. This could be achieved by hiding all the methods behind a single object, accessible using only one method, e.g. `pundit`.

This has its big downsides and probably is considered a non-Rails way. I would personally avoid that since you can always use Pundit without including it to your controller.

Burgestrand:
Merged #621 by means of #700. Closing this!

PR #570, opened 2018-12-11
Removes Error class from Pundit's global namespace

### LINKED ISSUE #542: "Error" namespace conflicts after including Pundit

tiagofsilva:
Fixes #542

This is a very simple way to solve #542 without changing the name of the Error class and without breaking the public API. 
This PR adds a new module `Pundit::Errors` that currently contains only `Error` class, thus avoiding the namespace to be polluted with a too common name, risking conflicts. The creation of a nested module for errors is good IMO because it helps usability in the future, i.e, opening the possibility of adding other errors to the private and even to the public API.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -8,18 +8,20 @@ require "active_support/core_ext/object/blank"
 require "active_support/core_ext/module/introspection"
 require "active_support/dependencies/autoload"
 
+# @api private
+module Pundit
+  module Errors
+    # Generic error that serves as base for all the other Pundit specific errors raised when
+    # authorization has failed
+    class Error < StandardError; end
+  end
+end
+
 # @api public
 module Pundit
   SUFFIX = "Policy".freeze
 
-  # @api private
-  module Generators; end
-
-  # @api private
-  class Error < StandardError; end
-
-  # Error that will be raised when authorization has failed
-  class NotAuthorizedError < Error
+  class NotAuthorizedError < Pundit::Errors::Error
     attr_reader :query, :record, :policy
 
     def initialize(options = {})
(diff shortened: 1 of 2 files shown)


kjleitz:
@tiagofsilva this looks okay, but duplicates https://github.com/varvet/pundit/pull/555 and results in the same overall behavior. The only difference between this one and mine is that this one results in:

```ruby
# api private
Pundit::Errors::Error # new namespace just for this one

# api public
Pundit::NotAuthorizedError # rest are in outer namespace (but inherit from the nested one)
Pundit::InvalidConstructorError
Pundit::AuthorizationNotPerformedError
Pundit::NotDefinedError
```

...whereas https://github.com/varvet/pundit/pull/555 results in:

```ruby
# api private
Pundit::PunditError # consistent namespace, doesn't cause the others to inherit from a "child"

# api public
Pundit::NotAuthorizedError
Pundit::InvalidConstructorError
Pundit::AuthorizationNotPerformedError
Pundit::NotDefinedError
```

It's mostly a semantic difference, as both are breaking changes _only_ if a user is inappropriately accessing a private part of the API in their project (i.e., `Pundit::Error`), but IMHO it makes more sense to have the "parent" error class be in the same scope (or above) as the classes which inherit from it, not to be nested _more deeply_ than the classes which inherit from it. Since our PRs accomplish the same goals, that would be my deciding factor, personally.

tiagofsilva:
@kjleitz Thanks for the heads up! They do both result in the same behaviour. Though I really don't like the repetition in the name``Pundit::PunditError`. None of the public error classes (`NotAuthorizedError`,  `InvalidConstructorError` and so on) start with "Pundit" prefix, which makes it inconsistent IMHO. 
I accept your argument that in my PR `Error` is more deeply nested than its children. That's a collateral effect of removing Error class from the global namespace.
But the advantage of this strategy is that it opens up the possibility to add other private error classes for now. Also, when it is time to refactor the Pundit module, nesting modules for errors, core and so on, this `Errors` module will already be in place (this enhancement is pointed out here: https://github.com/varvet/pundit/issues/523). I just think it's more of a "complete" solution. But I'm open to change my mind for sure. =)

Linuus:
What do you think about https://github.com/varvet/pundit/pull/590?

tiagofsilva:
@Linuus Yup, this is the ideal fix IMO.

PR #571, opened 2018-12-14
Updating `policy` in view example.

laurosilvacom:
Updating policy through the `policy`method in the view from `<% if policy(@post).update? %>` to `<% if policy(Post).update? %>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,9 +179,10 @@ method in both the view and controller. This is especially useful for
 conditionally showing links or buttons in the view:
 
 ``` erb
-<% if policy(@post).update? %>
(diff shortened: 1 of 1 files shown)


Linuus:
Why would you want to do this change?

You should pass the `@post` to the policy function so you can authorize that you may update that specific post. If you pass `Post` you can only authorize that the user can update posts in general.

laurosilvacom:
Honestly, that's the only way that I found for Post to parse the authorization. @post did not work on my end, I kept getting an error.

Linuus:
Please open an issue with the code you’re using and I’m sure someone will help out :)

PR #586, opened 2019-03-17
Only new should be except from scoping

MyklClason:
Any time a lookup is likely be done (IE: Model.find(id)), this should also be scoped otherwise they may be able to lookup records they don't strictly have access to.

It isn't sufficient to only scope the index page.

Edit: Actually this should be except new and create.


Linuus:
Hi!

This is just an example in the readme. You may add any exceptions/inclusions you want. I think only index is the most common case so I think the readme should stay as it is.

PR #590, opened 2019-04-17
Define Pundit::Error using compact style.

Linuus:
When defining the class with:

```ruby
module Pundit
  class Error
  end
end
```

the `Error` class is mixed in to the controller when using
`include Pundit`. This can cause issues since `Error` is a
very common class name. Using the compact style:

```ruby
class Pundit::Error
```

we avoid this, since it only creates the constant `Pundit::Error`.

Solves #542
Closes #555 #570

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -8,15 +8,17 @@ require "active_support/core_ext/object/blank"
 require "active_support/core_ext/module/introspection"
 require "active_support/dependencies/autoload"
 
+# @api private
+# To avoid name clashes with common Error naming when mixing in Pundit,
+# keep it here with compact class style definition.
+class Pundit::Error < StandardError; end # rubocop:disable Style/ClassAndModuleChildren
+
 # @api public
 module Pundit
   SUFFIX = "Policy".freeze
 
   # @api private
   module Generators; end
-
-  # @api private
-  class Error < StandardError; end
 
   # Error that will be raised when authorization has failed
   class NotAuthorizedError < Error


jensljungblad:
Yes!

jensljungblad:
I mean the same can be said for the other errors, even though name clashes wouldn't be as common. Why does it make sense to mixin an `InvalidConstructorError` into your controller?

Linuus:
> I mean the same can be said for the other errors, even though name clashes wouldn't be as common. Why does it make sense to mixin an `InvalidConstructorError` into your controller?

@jensljungblad Yeah, I agree. The reason I only did the `Error` case is because it's the most likely one to cause issues and it's marked as `@api private` for some reason :) That means we can fix the `Error` one without, hopefully, too many issues. If we move the other ones we need to bump the major version as it's a breaking change I'm afraid :(

jensljungblad:
Right, you mean if some people catch `NotAuthorizedError` instead of `Pundit::NotAuthorizedError`..

Linuus:
> Right, you mean if some people catch `NotAuthorizedError` instead of `Pundit::NotAuthorizedError`..
> 
> Well, I'm for doing it in the next major release then :)

Exactly :) My thought was doing this now and then fix everything in the next major release. I have some bigger structural changes I want to do anyway :)

kjleitz:
Nice! This is much better.

dgmstuart:
@Linuus just to check that I understand correctly: it sounds like this one can be merged and released now, and it would be nice to do change the other errors as well, but that will require a major version update?

Linuus:
IMO yes.

dgmstuart:
Rebasing...

lib/pundit.rb
```diff
@@ -8,16 +8,18 @@
...
   SUFFIX = "Policy".freeze
 
   # @api private
   module Generators; end
 
-  # @api private
-  class Error < StandardError; end
-
   # Error that will be raised when authorization has failed
   class NotAuthorizedError < Error
```
bf4: shouldn't this then be `< ::Pundit::Error` ?

dgmstuart: @bf4 Thanks for the comments - can you expand on this?

This does seem to inherit from the correct `Error` class if that's the concern:

```
pry(main)> require 'pundit'
=> true
pry(main)> Pundit::NotAuthorizedError.ancestors
=> [Pundit::NotAuthorizedError,
 Pundit::Error,
 StandardError,
 Exception,
 Object,
 JSON::Ext::Generator::GeneratorMethods::Object,
 PP::ObjectMixin,
 Kernel,
 BasicObject]
```

Issue #544, opened 2018-08-20
permitted_attributes_for_#{action} not working?


nesrual:
RoR 5.2.1
    Pundit 2.0.0
    Ruby 2.5.1

When updating an object I get:

    NoMethodError (undefined method `permitted_attributes' for #<AccountPolicy:0x00007f3920c163e8>
    Did you mean?  permitted_attributes_for_edit):

It works fine if I define `permitted_attributes` but fails if I only have `permitted_attributes_for_edit` and `permitted_attributes_for_create` defined in the policy file.

Linuus:
Can you show the code? What controller action are you in and what does your policy look like?

nesrual:
In the controller I fetch the attributes from a JSON_API request:

```  
def account_params
 ActiveModelSerializers::Deserialization.jsonapi_parse(params).slice(*policy(@account).permitted_attributes)
end
```

The policy looks like this:

```
  # def permitted_attributes
  #   %i[name account_type max_users active parent_id admin_user_ids editor_user_ids reader_user_ids]
  # end

  def permitted_attributes_for_edit
    %i[name account_type max_users active parent_id admin_user_ids editor_user_ids reader_user_ids]
  end

  def permitted_attributes_for_create
    %i[name account_type max_users active parent_id admin_user_ids editor_user_ids reader_user_ids]
  end
```

And it throws the following error:

```
NoMethodError (undefined method `permitted_attributes' for #<AccountPolicy:0x0000564891b027e8>
Did you mean?  permitted_attributes_for_edit):
```

nesrual:
The accounts controller looks like this:

```
  after_action :verify_authorized

  # POST /accounts
  def create
    @account = Account.new
    @account.update_attributes(account_params)
    authorize @account
    if @account.save
      render json: @account, status: :created, location: @account
    else
      render json: @account, status: :unprocessable_entity, adapter: :json_api, serializer: ActiveModel::Serializer::ErrorSerializer
    end
  end

  private

  # Only allow a trusted parameter "white list" through.
  def account_params
    ActiveModelSerializers::Deserialization.jsonapi_parse(params).slice(*policy(@account).permitted_attributes)
  end

  def pundit_user
    @current_user
  end
```

Linuus:
You are calling `permitted_attributes` on the policy so of course you must define it.

`policy(@account).permitted_attributes` here, `policy(@account)` returns the policy.

nesrual:
[...]

PR #591, opened 2019-04-18
Feature/implicitly change action name for crud

yfxie:
For most simple Rails app have actions named new, create, edit and update.
action new and create should share permitted parameters. so do action edit and update.
implicitly changing action name should help people resolve issues like https://github.com/varvet/pundit/issues/544. Otherwise, it's very confusing to choose name between edit and update and the `permitted_attributes` helper does not help 😕

For example
```
class PostPolicy
  def permitted_attributes_for_update
    [:title, :content]
  end
end

class Controller
  def edit
    permitted_attributes(@post)
  end
  def update
    permitted_attributes(@post)
  end
end
```
`permitted_attributes(@post)` should be expected to be same for the action edit and update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,13 +707,15 @@ class PostPolicy < ApplicationPolicy
     [:title, :body]
   end
 
-  def permitted_attributes_for_edit
+  def permitted_attributes_for_update
     [:body]
   end
 end
 ```
 
 If you have defined an action-specific method on your policy for the current action, the `permitted_attributes` helper will call it instead of calling `permitted_attributes` on your controller.
+
+For CRUD convention, `permitted_attributes` will use `permitted_attributes_for_create` for actions `new` and `create`, and `permitted_attributes_for_update` for actions `edit` ad `update`.
 
 If you need to fetch parameters based on namespaces different from the suggested one, override the below method, in your controller, and return an instance of `ActionController::Parameters`.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -707,14 +707,16 @@ class PostPolicy < ApplicationPolicy
...
-  def permitted_attributes_for_edit
+  def permitted_attributes_for_update
     [:body]
   end
 end
 ```
 
 If you have defined an action-specific method on your policy for the current action, the `permitted_attributes` helper will call it instead of calling `permitted_attributes` on your controller.
 
+For CRUD convention, `permitted_attributes` will use `permitted_attributes_for_create` for actions `new` and `create`, and `permitted_attributes_for_update` for actions `edit` ad `update`.
```
timkrins: Spelling of 'and' missing an 'n'

```suggestion
For CRUD convention, `permitted_attributes` will use `permitted_attributes_for_create` for actions `new` and `create`, and `permitted_attributes_for_update` for actions `edit` and `update`.
```

timkrins:
Just a note: this would be a breaking change for anyone who currently uses `permitted_attributes_for_edit` and `permitted_attributes_for_new`

Linuus:
Hi!

Thank you for your contribution 🙌

However, I don’t like this implicit behavior, I like that it’s explicit :)
If you want them to be the same, make an alias or something in your application policy. That’s easy enough for those who want it and no magic for everybody who doesn’t :)

ce07c3:
I'd even argue that the default `permitted_attributes` can lead to issues.

Issue #594, opened 2019-05-23
RuboCop 0.70.0 has new rule names


olleolleolle:
https://travis-ci.org/varvet/pundit/jobs/536151698 The linting job failed with:

```
.rubocop.yml:81: `Style/TrivialAccessors` is concealed by line 96
Error: The `Layout/IndentHash` cop has been renamed to `Layout/IndentFirstHashElement`.
(obsolete configuration found in .rubocop.yml, please update it)
```

rubyrider:
But Rubocop after 0.69 ended the support for Ruby 2.2. So I prefer locking the rubocop version to 0.68 and fix the existing redundant configuration until the core team decided to end the support for ruby 2.2 as the minimum ruby version.

I have created a pull request considering the above context.
 https://github.com/varvet/pundit/pull/598

olleolleolle:
I'd say, lock to an earlier RuboCop, make it pass.

Then, also, we could have a Ruby version support EOL policy discussion. What's the oldest Ruby we want to support?

rubyrider:
@olleolleolle 29 days ago I have sent an pull request which fix the rubocop version and keep old ruby supported by just updated few rubocop rules!  Would you be interested to review that ?
PR: https://github.com/varvet/pundit/pull/598

Looking forward for your feedback!

dgmstuart:
This specific issue has been fixed by #605 which removed both of these rules from the rubocop config entirely, and locked it to a passing version.

PR #597, opened 2019-06-07
Fix `rubocop` obsolete configuration

guilhermecortes:
This PR fixes the issue with the `rubocop` obsolete configuration
which was renamed from `Layout/IndentHash` to
`Layout/IndentFirstHashElement`.

With that change, some other changes were needed:
- update the `TargetRubyVersion` to 2.3.
- fix rubocop violations
- disable safe navigation with `Style/SafeNavigation`

This PR also removes the `TrivialAccessors` rule because it seems
unnecessary, see: https://github.com/varvet/pundit/pull/597#discussion_r312828652

Closes: #594

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -78,9 +78,6 @@ Style/SignalException:
 Layout/IndentationWidth:
   Enabled: false
 
-Style/TrivialAccessors:
-  ExactNameMatch: true
-
 Layout/EndAlignment:
   EnforcedStyleAlignWith: variable
 
@@ -95,6 +92,7 @@ Style/SpecialGlobalVars:
 
 Style/TrivialAccessors:
   Enabled: false
+  ExactNameMatch: true
 
 Layout/IndentFirstHashElement:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -90,10 +90,6 @@ Lint/HandleExceptions:
 Style/SpecialGlobalVars:
   Enabled: false
 
-Style/TrivialAccessors:
-  Enabled: false
-  ExactNameMatch: true
-
 Layout/IndentFirstHashElement:
   Enabled: false
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 AllCops:
   DisplayCopNames: true
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.3
   Exclude:
     - "gemfiles/**/*"
     - "vendor/**/*"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -101,3 +101,6 @@ Layout/IndentFirstHashElement:
 
 Style/DoubleNegation:
   Enabled: false
+
+Style/SafeNavigation:
+  Enabled: false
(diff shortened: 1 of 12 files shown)


rubyrider:
This update needs to disable safe navigation then. I prefer to lock the rubocop version unless the ruby supports end for 2.2.  

I have created a pull request considering the above context.

rubyrider:
If you use safe navigation then Pundit will be broken for the ruby version less than 2.3.0.

guilhermecortes:
Yep. I've disabled the safe navigation.

rubyrider:
Thanks!

guilhermecortes:
I'm unsure what is the current issue with the CI though.

dgmstuart:
@guilhermecortes I think it's a good idea to lock the rubocop version, right? I have some time to work on Pundit now so I can do that in a separate PR

guilhermecortes:
Cool, thanks.

dgmstuart:
@guilhermecortes would you mind trying to rebase on top of the latest master, or let me know if you'd prefer if I take this forward. Maybe worth also having a scan of https://github.com/varvet/pundit/pull/598 to see if there's any overlap

.rubocop.yml
```diff
@@ -95,6 +92,7 @@ Style/SpecialGlobalVars:
 
 Style/TrivialAccessors:
   Enabled: false
+  ExactNameMatch: true
```
dgmstuart: Hmmm. The first declaration expected it to be on, and this second one expected turned it off. If it's disabled like this then is there any value in specifying configuration?

Separate point: do we definitely want it to be switched off?

dgmstuart:
@guilhermecortes This is failing because it's still using rubocop 0.57.2, which doesn't have `Layout/IndentFirstHashElement`. 

Could you please change the version to 0.74.0 (the latest version) in the gemspec - that should sort it out. I don't know what the issue is with the JQuery failure yet - I'll look into that.

.rubocop.yml
dgmstuart: Haha - looks like in the initial commit which added rubocop we have both the rule saying "exact name match" and a rule saying "Enabled: false" - looks like it was just copy-pasted from another project: I think we can just delete this rule, since the code seems fine without it.

dgmstuart:
Hmm - I reran the failing JRuby build and it works now. Maybe just a momentary Travis glitch 🤷‍♂

guilhermecortes:
> Hmm - I reran the failing JRuby build and it works now. Maybe just a momentary Travis glitch 🤷‍♂

🤷‍♂ 

>@guilhermecortes This is failing because it's still using rubocop 0.57.2, which doesn't have Layout/IndentFirstHashElement.
Could you please change the version to 0.74.0 (the latest version) in the gemspec - that should sort it out. 

Will do!

.rubocop.yml
guilhermecortes: Ok, I'll give it a try.

dgmstuart:
@guilhermecortes Ok one more try: we need version 0.68.1 in the gemspec - that's the last version that supports ruby 2.1 and 2.2

guilhermecortes:
Seems like it still failing with ruby 2.1

dgmstuart:
@guilhermecortes Yeah - unfortunately there is no version of rubocop which works with both this PR and with Ruby 2.1. I'd like to get some of these changes in before making a breaking change to remove support for the old ruby versions - I'll create a new PR for that, but I think this one should be closed for now.

Thanks for all your help!

PR #598, opened 2019-07-07
Update Rubocop rules and version

rubyrider:
This pull request fix and update the following items:

- [x] Sets the `Rubocop` version to 0.68.1 as after 0.69 targetted ruby version 2.2 is not supported. 

- [x] Removes unnecessary following rules as the rule is already disabled.
```ruby
Style/TrivialAccessors:
  ExactNameMatch: true
```
- [x]  Rename 
```ruby
Layout/IndentHash:
  Enabled: false
```
To:
```ruby
Layout/IndentFirstHashElement:
  Enabled: false
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,32 +30,8 @@ Metrics/CyclomaticComplexity:
 Metrics/PerceivedComplexity:
   Enabled: false
 
-Style/StructInheritance:
-  Enabled: false
-
 Layout/AlignParameters:
   EnforcedStyle: with_fixed_indentation
-
-Style/StringLiterals:
-  EnforcedStyle: double_quotes
-
-Style/StringLiteralsInInterpolation:
-  EnforcedStyle: double_quotes
-
-Layout/ClosingParenthesisIndentation:
-  Enabled: false
-
-Style/OneLineConditional:
-  Enabled: false
-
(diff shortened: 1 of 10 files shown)


rubyrider:
Thanks for your approval @olleolleolle .
Let me know if you require any further changes! I am happy to work on this.

olleolleolle:
One issue to solve is the 2.1 being a failure now, because of the Rubocop presence. 

Perhaps it's possible to create a separate Gemfile for 2.1, pointing that out in the Travis config using the "gemfile:" directive? And, well, ensure that for 2.1, the Rubocop tool is not run. 

Going green is the goal. Can be a bit of hassle, and perhaps the project (of which I'm not a part) decides on a Ruby version support policy and path forward before green on this branch has been achieved. But, it's possible, at least, to get to green.

rubyrider:
@olleolleolle let me check. Thank you for the concern.

dgmstuart:
Hi @rubyrider would you mind trying to rebase your PR branch on top of the latest `master`? It should build successfully now. Note that the Gem specification has been moved to the gemspec so you can do the rubocop version change there instead.

rubyrider:
@dgmstuart yes I am going to rebase it now.

PR #605, opened 2019-08-12
Rubocop Housekeeping

dgmstuart:
The intention of this PR is to clean up the Rubocop config, and make it forwards-compatible as much as possible without dropping support for ruby 2.1 and 2.2.

There are a lot of things in the rubocop config which don't really need to be
there: a number seem to have been copied from some other project. 

Where there are no issues, or the issues can be easily fixed to be in line with the rubocop
defaults, then it feels like a good idea to remove these exceptions, at least
in part because it insulates us from future changes to the cop name or options.

### DIFF
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -518,11 +518,13 @@ describe Pundit do
 
   describe "#permitted_attributes" do
     it "checks policy for permitted attributes" do
-      params = ActionController::Parameters.new(post: {
-        title: "Hello",
-        votes: 5,
-        admin: true
-      })
+      params = ActionController::Parameters.new(
+        post: {
+          title: "Hello",
+          votes: 5,
+          admin: true
+        }
+      )
 
       action = "update"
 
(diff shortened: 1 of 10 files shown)


dgmstuart:
@rubyrider @guilhermecortes @olleolleolle you've all been involved in the rubocop discussion - feel free to comment if you have a moment - otherwise I'll merge this later this week.

spec/pundit_spec.rb
```diff
@@ -518,11 +520,13 @@
...
-      params = ActionController::Parameters.new(post: {
-        title: "Hello",
-        votes: 5,
-        admin: true
-      })
+      params = ActionController::Parameters.new(
+        post: {
+          title: "Hello",
+          votes: 5,
+          admin: true
```
guilhermecortes: What do you think about adding the `TrailingCommaInArguments` rule?
It makes the diff smaller when adding new things.

guilhermecortes: This should also be done in a different PR though.

dgmstuart: I do like that rule, particularly on projects which have a lot of updates, but I also like the idea of keeping the rubocop config as small as possible, so that hopefully it breaks less often. This project doesn't often have updates and can go for long periods without updates so maybe that makes a difference?

dgmstuart: I'm open to a separate PR

rubyrider: I think this should be done in separate PR as well.

rubyrider:
Looks good!

PR #608, opened 2019-08-19
CI: Name specific steps in Travis CI matrix

olleolleolle:
This PR uses the `name` feature for CI matrix steps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,19 @@ before_install:
 
 matrix:
   include:
-    - rvm: 2.5.3 # Pre-installed Ruby version
+    - name: "RuboCop lint on pre-installed Ruby version"
+      rvm: 2.5.3 # Pre-installed Ruby version
       script: bundle exec rake rubocop # ONLY lint once, first
     - rvm: 2.3.5
     - rvm: 2.4.6
     - rvm: 2.5.5
     - rvm: 2.6.3
-    - rvm: jruby-9.1.8.0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,16 +4,18 @@ before_install:
...
+    - name: "jruby-9.1.8.0 on OpenJDK 8"
+      rvm: jruby-9.1.8.0
       jdk: openjdk8
     - rvm: jruby-9.2.8.0
-      env:
-        - JRUBY_OPTS="--debug"
+
+env:
+  global:
+    - JRUBY_OPTS="--debug"
```
dgmstuart: This only applies to the JRUBY builds - but does this apply it to all builds? I guess it isn't harmful if it's present, but does it feel odd to be setting it for all of the MRI builds?

olleolleolle: You are correct: This configuration sets it globally, to all matrix elements.

Most don't care about the presence of this environment variable.

The test suite is using `$JRUBY_OPTS` as a non-specific "just use the convenient --debug settings", as opposed to a matrix-selecting environment variable (e.g. ISOLATED=true, ISOLATED=false in the example at https://docs.travis-ci.com/user/build-matrix/ )

olleolleolle: Would you like me to exclude this change from this PR?

dgmstuart: @olleolleolle I'm not 100% understanding, but maybe it's simplest to exclude that change for now, if it's not directly related to the `name` change?

PR #611, opened 2019-08-19
Only install bundler on needy version

dgmstuart:
It seems that on the preinstalled version of ruby we (sometimes?) don't
get a preinstalled bundler for free, but on rubies which we build and
install ourselves we do.

By not installing before every single build we save ourselves a few
seconds in the overall build.


dgmstuart:
@olleolleolle as per your suggestion

PR #616, opened 2019-08-23
Scopes have private attr_readers

dgmstuart:
- Generate application scope with `protected` attr readers
- Update documentation to suggest that other scopes should use `private`

There doesn't seem to be a use case for accessing the initialisation
parameters on scopes. In general keeping the public interface of classes
as small as possible can help to keep code decoupled.

The application scope readers need to be protected rather than private  since
other scopes are expected to inherit from it

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,9 +37,6 @@ Layout/CaseIndentation:
     - end
   IndentOneStep: true
 
-Layout/AccessModifierIndentation:
-  EnforcedStyle: outdent
-
 Layout/EndAlignment:
   EnforcedStyleAlignWith: variable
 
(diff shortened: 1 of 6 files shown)


Linuus:
> The application scope readers need to be protected rather than private since
other scopes are expected to inherit from it

That's not how private/protected work with inheritance in Ruby :) Inheriting classes can access private stuff from the parent class.

Protected is used when another instance of the same class needs access but not instances of other classes.

dgmstuart:
@Linuus huh - I looked it up before writing this, but I clearly I misunderstood:

```ruby
class A
  private

  def x
    puts "a useful value"
  end

  protected

  def y
    puts "another value"
  end
end

class B < A
  def bx
    x
  end

  def by
    y
  end
end

A.new.x # => private method `x' called for...
A.new.y # => protected method `y' called for...
B.new.bx # => "a useful value"
B.new.by # => "another value"
```

dgmstuart:
@Linuus I've squashed in a change to make it private and update the commit message

PR #619, opened 2019-09-05
Prefer single-quoted strings

geoom:
Prefer single-quoted strings when you don't need string interpolation or special symbols


Linuus:
No. There’s no reason to use single quotes strings.

PR #621, opened 2019-10-21
Extract Pundit controller extension module

gogiel:
Related Issue: https://github.com/varvet/pundit/issues/523
Fixes https://github.com/varvet/pundit/issues/542

Including top level `Pundit` module has huge side effects. It not only includes methods defined in `Pundit` module, but also adds all internal classes and modules to the scope. And not only the modules that are defined in `pundit.rb` but literally everything! In your controller you can access constants like `VERSION`, `SUFFIX`, `Error` or even `RSpec` from `lib/pundit/rspec.rb`.

I'm looking for a name better than `Extension`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@
 - Dropped support for Ruby end-of-life versions: 2.3 (#633)
 - Dropped support for RSpec 2 (#615)
 
+### Deprecated
+
+- Deprecate `include Pundit` in favor of `include Pundit::Authorization` (#621)
+
 ## 2.1.0 (2019-08-14)
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ Sponsored by:
 gem "pundit"
 ```
 
-Include Pundit in your application controller:
+Include `Pundit::Authorization` in your application controller:
 
 ``` ruby
 class ApplicationController < ActionController::Base
-  include Pundit
+  include Pundit::Authorization
   protect_from_forgery
 end
 ```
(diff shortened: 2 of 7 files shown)


dgmstuart:
@Linuus I feel like this has come up before - do you have any thoughts?

Linuus:
@dgmstuart Yep. I started this on the pundit-3 branch a long time ago: https://github.com/varvet/pundit/tree/pundit-3

dgmstuart:
@Linuus if it's done as a deprecation like this I think it could be a minor version. Do you agree?

@gogiel I had a look at @Linuus's branch, which called the module `Pundit::Authorization` - how do you feel about that?

dgmstuart:
(Though actually we're probably looking at doing a major version anyway since we dropped some compatibilities)

@gogiel would you mind adding a line to the changelog under `unreleased` as `### Deprecated`

dgmstuart:
@gogiel also I'm afraid you might have a bad time rebasing since there have been a number of added and changed tests. Maybe worth re-doing part of it from scratch I'm afraid 😬

dgmstuart:
@Linuus since this is a significant change it would be great to have a second pair of 👀 if you have time. The PR seems fairly straightforward.

dgmstuart:
@gogiel Great work by the way - feels thorough 👍

uxxman:
Any updates on merging this?

Burgestrand:
Hi! I took the liberty of rebasing this on the latest code and merged it in #700, so I'll close this PR. Thanks for this :)

sekrett:
Many thanks for this. But you added deprecation warning to Changelog for version 2.1.1, which was released 5 months ago, it should be moved to unreleased section. @Burgestrand

Burgestrand:
> Many thanks for this. But you added deprecation warning to Changelog for version 2.1.1, which was released 5 months ago, it should be moved to unreleased section. @Burgestrand

Thank you, that is indeed the wrong place!

PR #623, opened 2019-10-28
Ability for rspec matcher to take additional args

dwfait:
Sometimes we need to add custom args to our policy methods, and would be
nice to use the built in rspec matchers to test these policies. This
allows arbitrary arguments to be passed in to the policy.


dwfait:
closing as upon review of the docs, pundit explicitly does not support this feature. The codebase I'm working on adds this, but looks like refactoring to either use the context or to have a separate model representing the join would be better solutions.

PR #624, opened 2019-10-29
Raise error when try to authorize nilclass

gabrielbaldao:
Can we add this change to when someone try to authorize some nilclass?


Linuus:
As you can see in the readme we explicitly want to allow nil

https://github.com/varvet/pundit/blob/78c509d9fb66061ac9d4ba82614988059e5ba64c/README.md#nilclasspolicy

PR #626, opened 2019-11-10
'.authorize' and '#authorize' return record even with passed record with namespace array

QWYNG:
Hi, Thank you for the great Gem!
This is a proposal and pull request at once.
'.authorize' and '#authorize' return passed record now.

```ruby
def show
  @post  = authorize Post.find(params[:id])
end
```

But When passing record with a namespace. return array

```ruby 
def show
  @post  = authorize [:admin, Post.find(params[:id])]
end

# @post is  [:admin, Post.find(params[:id])]
```
These methods are expected to return record not namespase array

When authorize override the helpers in AdminController to automatically apply the namespacing, This change will be very useful.

```ruby
    super([:admin, record], query)
  end
end

class Admin::PostController < AdminController
  def show
    post = authorize Post.find(params[:id])
    # We don't need `authorize(post)` line!
  end
end
```

If there is a place to fix, please let me know.
best regards.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -71,7 +71,7 @@ module Pundit
 
       raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
-      record
+      record.is_a?(Array) ? record.last : record
     end
 
     # Retrieves the policy scope for the given record.
@@ -222,7 +222,7 @@ protected
 
     raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
-    record
+    record.is_a?(Array) ? record.last : record
   end
 
   # Allow this action not to perform authorization.
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -25,6 +25,14 @@ describe Pundit do
       expect(Pundit.authorize(user, post, :update?)).to be_truthy
     end
 
+    it "returns the record on successful authorization" do
+      expect(Pundit.authorize(user, post, :update?)).to be(post)
+    end
+
+    it "returns the record when passed record with namespace " do
+      expect(Pundit.authorize(user, [:project, comment], :update?)).to be(comment)
+    end
+
     it "can be given a different policy class" do
       expect(Pundit.authorize(user, post, :create?, policy_class: PublicationPolicy)).to be_truthy
     end
@@ -411,6 +419,10 @@ describe Pundit do
 
     it "returns the record on successful authorization" do
       expect(controller.authorize(post)).to be(post)
+    end
+
+    it "returns the record when passed record with namespace " do
+      expect(Pundit.authorize(user, [:project, comment], :update?)).to be(comment)
     end
 
     it "can be given a different permission to check" do
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -37,16 +37,8 @@ describe Pundit do
       expect(Pundit.authorize(user, [:project, :admin, comment], :update?)).to eq(comment)
     end
 
-    it "can use headless policy" do
-      expect(Pundit.authorize(user, :publication, :create?)).to be_truthy
-    end
-
     it "returns the policy name symbol when passed record with headless policy" do
       expect(Pundit.authorize(user, :publication, :create?)).to eq(:publication)
-    end
-
-    it "can use without a particular instance" do
-      expect(Pundit.authorize(user, Post, :show?)).to be_truthy
     end
 
     it "returns the class when passed record not a particular instance" do
@@ -449,16 +441,8 @@ describe Pundit do
       expect(controller.authorize([:project, :admin, comment], :update?)).to eq(comment)
     end
 
-    it "can use headless policy" do
-      expect(controller.authorize(:publication, :create?)).to be_truthy
-    end
-
     it "returns the policy name symbol when passed record with headless policy" do
       expect(controller.authorize(:publication, :create?)).to eq(:publication)
-    end
-
-    it "can use without a particular instance" do
-      expect(controller.authorize(Post, :show?)).to be_truthy
     end
 
     it "returns the class when passed record not a particular instance" do
--- a/spec/pundit_spec.rb
+++ b/spec/pundit_spec.rb
@@ -33,6 +33,26 @@ describe Pundit do
       expect(Pundit.authorize(user, [:project, comment], :update?)).to be(comment)
     end
 
+    it "returns the record when passed record with nested namespace " do
+      expect(Pundit.authorize(user, [:project, :admin, comment], :update?)).to be(comment)
+    end
+
+    it "can use headless policy" do
+      expect(Pundit.authorize(user, :publication, :create?)).to be_truthy
+    end
+
+    it "returns the policy name symbol when passed record with headless policy" do
+      expect(Pundit.authorize(user, :publication, :create?)).to be(:publication)
+    end
+
+    it "can use without a particular instance" do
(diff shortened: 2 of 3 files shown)


spec/pundit_spec.rb
```diff
@@ -25,6 +25,14 @@
       expect(Pundit.authorize(user, post, :update?)).to be_truthy
     end
 
+    it "returns the record on successful authorization" do
+      expect(Pundit.authorize(user, post, :update?)).to be(post)
+    end
+
+    it "returns the record when passed record with namespace " do
+      expect(Pundit.authorize(user, [:project, comment], :update?)).to be(comment)
+    end
```
dgmstuart: For completeness, is it worth adding another spec for the nested namespace case (`[:foo, :bar, Post]`)?https://github.com/varvet/pundit#policy-namespacing

spec/pundit_spec.rb
```diff
@@ -413,6 +421,10 @@
       expect(controller.authorize(post)).to be(post)
     end
 
+    it "returns the record when passed record with namespace " do
+      expect(Pundit.authorize(user, [:project, comment], :update?)).to be(comment)
+    end
```
dgmstuart: This looks identical to the `.authorize` spec. Should this be `controller.authorize` instead?

spec/pundit_spec.rb
```diff
@@ -25,6 +25,14 @@
       expect(Pundit.authorize(user, post, :update?)).to be_truthy
     end
 
+    it "returns the record on successful authorization" do
+      expect(Pundit.authorize(user, post, :update?)).to be(post)
+    end
```
dgmstuart: What is the return value in the case of a headless policy? https://github.com/varvet/pundit#headless-policies Is it worth having a test for that here?

Also in the case where there is no instance (`authorize Post`): https://github.com/varvet/pundit#policies

Ideally there would already be test cases for these and it shouldn't be your responsibility to implement them, but since we're changing the behaviour, we should probably make sure that we're not breaking any existing behaviour.

lib/pundit.rb
```diff
@@ -71,7 +71,7 @@ def authorize(user, record, query, policy_class: nil)
 
       raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
-      record
+      record.is_a?(Array) ? record.last : record
```
dgmstuart: Another implementation might be `Array(record).last`, but maybe that's harder to understand?

dgmstuart:
While I think it's unlikely that anyone is relying on getting the namespace array back from the authorize method, technically this is a breaking change: I don't think there's a clear expectation set that we always return the model instance, since the README does say "authorize returns the object passed to it", which I'd argue could mean either "the instance" or "the array". 

Not sure how seriously we need to take that 🤷‍♂

spec/pundit_spec.rb
QWYNG: I agree with what you are saying.

lib/pundit.rb
QWYNG: `Kernel.Array` can cause things like:
```ruby 
hash = { name: `Bob`, age: 40 }

Array(hash).last
#=> [:age, 40]
```

I want to reduce the possibility of modifying the record itself...
What do you think?

QWYNG:
@dgmstuart 
Thank you for your comments!
I'll fix spec!
I think that this is a breaking change too.
But I think that most users would expect the only record itself to return in this case.

lib/pundit.rb
dgmstuart: Yeah you're right. I guess I have a general discomfort around typechecks, but doing some source-diving it seems like the array form is how Ruby looks up namespaces, so I guess that's just how it is 🤷‍♂

QWYNG:
@dgmstuart 
Thank you for your review.
I fixed spec.
I'm really grateful as a pundit user for your interaction even with this PR doesn't marge.

spec/pundit_spec.rb
```diff
@@ -33,6 +33,26 @@
...
+    it "returns the policy name symbol when passed record with headless policy" do
+      expect(Pundit.authorize(user, :publication, :create?)).to be(:publication)
+    end
+
+    it "can use without a particular instance" do
+      expect(Pundit.authorize(user, Post, :show?)).to be_truthy
+    end
+
+    it "returns the class when passed record not a particular instance" do
+      expect(Pundit.authorize(user, Post, :show?)).to be(Post)
```
dgmstuart: `be` feels unusual to me - would `eq?` be more expressive/conventional? 
The same for "comment" above

spec/pundit_spec.rb
```diff
@@ -33,6 +33,26 @@
...
+    it "can use headless policy" do
+      expect(Pundit.authorize(user, :publication, :create?)).to be_truthy
+    end
+
+    it "returns the policy name symbol when passed record with headless policy" do
+      expect(Pundit.authorize(user, :publication, :create?)).to be(:publication)
+    end
+
+    it "can use without a particular instance" do
+      expect(Pundit.authorize(user, Post, :show?)).to be_truthy
```
dgmstuart: This feels like quite a weak assertion - is truthiness a property we care about? 

Would it be better to make this an assertion about the return value (`"returns the... "`) and to expect the specific value we get in this case? (is it `Post`?)

spec/pundit_spec.rb
QWYNG: OK!

spec/pundit_spec.rb
QWYNG: I see.

QWYNG: But we have not 'can use without a particular instance' spec and 'can use headless policy'.
May I add it at this time?

dgmstuart: The thing for me is that I'm not sure what "can use" (or "can be used") is intending to test.

For me there are usually two types of test, which correspond to the two kinds of property which we care about our code having:
1. What's the return value (if we care)
2. What side effects happen (if any)

If we have one of these tests for a particular scenario/context then we're also implicitly testing that we "can use" the method in that context, since we're using the code. If we don't care about either 1 or 2, then probably the code doesn't need to exist 😉 

Am I making sense, or am I missing the point?

QWYNG: Thanks! It became clear.
Certainly, 1 is testing that we "can use" the method.

Linuus:
I would definitely say this is a bug 🐛

QWYNG:
@dgmstuart 
Thank you for your review!
fixed again

spec/pundit_spec.rb
dgmstuart: Cool, in that case I suggest it should be the same shape as the other specs which test a return value:
title: `it "returns..." do`
expectation: `expect(Pundit.authorize(...).to eq(<specific value>)`

QWYNG: Yes
I left only test a return value.

spec/pundit_spec.rb
```diff
@@ -37,18 +37,10 @@
...
-      expect(Pundit.authorize(user, :publication, :create?)).to be_truthy
-    end
-
     it "returns the policy name symbol when passed record with headless policy" do
       expect(Pundit.authorize(user, :publication, :create?)).to eq(:publication)
     end
 
-    it "can use without a particular instance" do
-      expect(Pundit.authorize(user, Post, :show?)).to be_truthy
-    end
```
dgmstuart: Ah I see - I hadn't read this correctly before. Yes, it's good to delete these specs, since they're entirely covered by the `it "returns..."` specs.

QWYNG:
@dgmstuart @Linuus 
I really appreciate both!
Thank you for the useful gem again!

Issue #603, opened 2019-08-07
Drop support for Ruby 2.1 and Ruby 2.2?


dgmstuart:
Ruby 2.1 is has been officially obsolete for over two years now, and no longer receives security updates: https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

Similarly Ruby 2.2 has been obsolete for over a year: https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

One impact of supporting older versions is that we can't keep our development dependencies up-to-date. This does add some friction for contributors - perhaps it's not on it's own a reason to drop support, but perhaps there are other reasons?

dgmstuart:
@Linuus @olleolleolle @cjse @thomasklemm @Burgestrand @jnicklas any opinions gratefully heard

Linuus:
IMO: Drop anything that has been EOL’d.

Burgestrand:
> IMO: Drop anything that has been EOL’d.

Agreed!

dgmstuart:
@Burgestrand Ah OK - you still have write access as a collaborator, should I remove you?

Burgestrand:
> @Burgestrand Ah OK - you still have write access as a collaborator, should I remove you?

Sure. If I were to contribute I'd send a PR either way :)

dgmstuart:
PR for the first part of this is ready for review: https://github.com/varvet/pundit/pull/604

dgmstuart:
#610 is a PR which commits the lockfile, but it has some issues around JRuby. I've tried a few things, but I'm a bit stuck. 

Perhaps it's not worth it to try and commit the lockfile (the main thing for reducing maintainer friction was to lock Rubocop to a specific version), but if anyone has a chance to check out that PR, it would be much appreciated.

Linuus:
FYI 2.3 is also EOL’d btw :)

PR #633, opened 2019-12-09
Stop building against ruby 2.3

dgmstuart:
It went end-of-life in March 2019:

https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/

Build against latest versions of other rubies


dgmstuart:
Addresses comment in https://github.com/varvet/pundit/issues/603

PR #635, opened 2019-12-31
Remove calling super in included

guppy0356:
This commit is to remove calling `super` in included method. Because `included` is callback to include or extend modules without calling `super`

### DIFF
--- a/lib/pundit/rspec.rb
+++ b/lib/pundit/rspec.rb
@@ -65,7 +65,6 @@ module Pundit
       def self.included(base)
         base.metadata[:type] = :policy
         base.extend Pundit::RSpec::DSL
-        super
       end
     end
   end


guppy0356:
Please respond if you are here.

Burgestrand:
Hi! I'm late to the party.

I'm not against this, but I haven't been able to find any documentation that says this is bad, or this is good.

I tried reading ruby source to see if there is a problem, but couldn't find any. I tried writing code examples that show a problem, but couldn't make a reasonable problem example.

When nothing is provably broken/improved my reaction is to not change anything. This doesn't seem broken to me, so there's nothing to fix by changing this?

guppy0356:
@Burgestrand 

Thank for your respond. I posted blog about this in Japanese.
https://qiita.com/guppy0356/items/4cc0025c314ca1404658

`super` is over-riding the behavior from a method provided in the ancestor chain. In this case, `super` is over-riding `Module.included`. `Module.included` does nothing. That's why I think this should be removed.

Burgestrand:
Thank you for the link! I read the blog post.

I see no problem, and no reason to change code.

Long answer:

I found implementation of `Module.included` in Ruby 3.1: https://github.com/ruby/ruby/blob/1db6249c66291ad2464f89768d24d850dbf98901/object.c#L1028-L1049

`Module.included` is defined with `rb_obj_dummy`, so there is no problem calling `Module.included`.

```ruby
static VALUE
rb_obj_dummy(void)
{
    return Qnil;
}
```

Even more, `Module.included` is always called on _every include_. I think not calling `super` breaks normal behavior, and it is wrong to forget.

```ruby
# frozen_string_literal: true

class Module
  def included(other)
    puts "#{self} (Module instance) included in #{other}"
  end
end

module A; end

module B
  def self.included(other)
    puts "#{self} (B) included in #{other}"
    super
  end
end

module C
  def self.included(other)
    puts "#{self} (C) included in #{other}"
  end
end

class X
  puts "A:"
  include A
  puts

  puts "B:"
  include B
  puts

  puts "C:"
  include C
end
```

Output:

- A is OK ✅
- B is OK ✅
- C is not normal ⚠️

> A:
> A (Module instance) included in X
> 
> B:
> B (B) included in X
> B (Module instance) included in X
> 
> C:
> C (C) included in X

guppy0356:
@Burgestrand 

Let's change our approach to the story. What is your purpose to call `super` in `self.included`. `Module#included` does nothing. What's the point of writing something that doesn't do anything?

Burgestrand:
It doesn't do nothing. It calls `Module#included`, which is the standard behavior. If `Module#included` is not called when including a module, that is not the normal behavior. 

I think `super` should be called, to keep normal behavior.

The second argument is that Pundit code has existed for a long time. We don't know who might be relying on this behavior. Most likely nobody, but there's a slight risk that removing it might break somebody's code. I think it's worth keeping around because 1) it doesn't do harm to keep it and 2) it _might_ do harm to remove it.

guppy0356:
@Burgestrand 

> Pundit code has existed for a long time. We don't know who might be relying on this behavior. Most likely nobody, but there's a slight risk that removing it might break somebody's code. I think it's worth keeping around because 1) it doesn't do harm to keep it and 2) it might do harm to remove it.

I am missing the point.  You are right, and you were right to close it.

PR #638, opened 2020-01-17
Fix issue in README for multiple error messages

### LINKED ISSUE #637: Multiple Error Messages per One Policy Action from README causes errors
I am following the instructions here: https://github.com/varvet/pundit#multiple-error-messages-per-one-policy-action

and after putting the rescue_from block in my ApplicationController, I get the following error:

```
/app/controllers/application_controller.rb:5: syntax error, unexpected ',', expecting end
...or] = message, scope: "pundit", default: :default
...                              ^
/app/controllers/application_controller.rb:73: syntax error, unexpected end, expecting end-of-input
```

Is there something I'm missing?

holyketzer:
Closes #637

### DIFF
--- a/README.md
+++ b/README.md
@@ -549,7 +549,7 @@ class ProjectPolicy < ApplicationPolicy
   def create?
     if user.has_paid_subscription?
       if user.project_limit_reached?
-        raise Pundit::NotAuthorizedError, reason: 'user.project_limit_reached', record: user, query: :create
+        raise Pundit::NotAuthorizedError, reason: 'user.project_limit_reached'
       else
         true
       end
@@ -563,7 +563,7 @@ end
 Then you can get this error message in exception handler:
 ```ruby
 rescue_from Pundit::NotAuthorizedError do |e|
-  flash[:error] = I18n.t("errors.#{e.reason}", scope: "pundit", default: e.message)
+  flash[:error] = e.reason ? I18n.t("errors.#{e.reason}", scope: "pundit") : e.message
   redirect_to(request.referrer || root_path)
 end
 ```
--- a/README.md
+++ b/README.md
@@ -549,7 +549,7 @@ class ProjectPolicy < ApplicationPolicy
   def create?
     if user.has_paid_subscription?
       if user.project_limit_reached?
-        raise Pundit::NotAuthorizedError, reason: 'user.project_limit_reached'
+        raise Pundit::NotAuthorizedError, reason: 'user.project_limit_reached', record: user, query: :create
       else
         true
       end
@@ -563,14 +563,7 @@ end
 Then you can get this error message in exception handler:
 ```ruby
 rescue_from Pundit::NotAuthorizedError do |e|
-  message = 
-    if e.reason 
-      I18n.t("pundit.errors.#{e.reason}", scope: "pundit", default: :default)
-    else 
-      e.message
-    end
-
-  flash[:error] = message
+  flash[:error] = I18n.t("errors.#{e.reason}", scope: "pundit", default: e.message)
   redirect_to(request.referrer || root_path)
 end
 ```
--- a/README.md
+++ b/README.md
@@ -563,8 +563,14 @@ end
 Then you can get this error message in exception handler:
 ```ruby
 rescue_from Pundit::NotAuthorizedError do |e|
-  message = e.reason ? I18n.t("pundit.errors.#{e.reason}") : e.message
-  flash[:error] = message, scope: "pundit", default: :default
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -563,8 +563,14 @@ end
 Then you can get this error message in exception handler:
 ```ruby
 rescue_from Pundit::NotAuthorizedError do |e|
-  message = e.reason ? I18n.t("pundit.errors.#{e.reason}") : e.message
-  flash[:error] = message, scope: "pundit", default: :default
+  message = 
+    if e.reason 
+      I18n.t("pundit.errors.#{e.reason}", scope: "pundit", default: :default)
```
dgmstuart: Is `scope` required? Isn't it covered by starting the key with `pundit`?

Is the `default:` an essential part of this approach? It seems a bit odd to specify a default and then to also provide a default in the `else` clause? I wonder if we can just remove this and go back to the ternary `if`?

holyketzer: Correct, I missed it. Just tested right now

holyketzer: I agree about this if, fixed it and tested

README.md
```diff
@@ -549,7 +549,7 @@ class ProjectPolicy < ApplicationPolicy
   def create?
     if user.has_paid_subscription?
       if user.project_limit_reached?
-        raise Pundit::NotAuthorizedError, reason: 'user.project_limit_reached'
+        raise Pundit::NotAuthorizedError, reason: 'user.project_limit_reached', record: user, query: :create
```
dgmstuart: Do we need this `record`/`query`? It seems like they're not used - just reason.

README.md
```diff
@@ -563,14 +563,7 @@ end
...
 rescue_from Pundit::NotAuthorizedError do |e|
-  message = 
-    if e.reason 
-      I18n.t("pundit.errors.#{e.reason}", scope: "pundit", default: :default)
-    else 
-      e.message
-    end
-
-  flash[:error] = message
+  flash[:error] = I18n.t("errors.#{e.reason}", scope: "pundit", default: e.message)
```
dgmstuart: Is there a special reason to use `scope` here? If not then can we keep this example as simple and conventional as possible by using the basic syntax: `pundit.errors...`

README.md
holyketzer: I suppose yes, because in this issue https://github.com/varvet/pundit/issues/637 is what exactly happened when no corresponding reason is found in `I18n` then use sees default message, and to show it correctly Pundit is required record and query else it will look like `not allowed to  this NilClass`

README.md
holyketzer: No there is no special reason for it. Here I just tried to follow current repo style, and that I found in above readme section https://github.com/varvet/pundit/blob/master/README.md#creating-custom-error-messages

dgmstuart: Ah I see: you're absolutely right. I don't like it, but it's consistent. Let's leave the `scope`.

README.md
dgmstuart: Sure, but why would we want to display a default message in this case?

If there's no corresponding reason in I18n, then in my opinion it should fail: If I set a `reason`, then I'm saying "I want a specific message to be displayed - not the default message", so if it's not possible to display a specific message (because the user forgot to add a key in the locales) then the behaviour should be to fail, not to cover up the fact that it failed.

I'm open to the idea that there's some valid use-case where a reason is specified, but the corresponding locale text isn't specified, but I can't think of one: can you?

I think the controller logic you had before was reasonable:

"If there's a reason set on the error, then look it up in I18n, otherwise show the `message` from the error"

holyketzer: It's reasonable, I agree

README.md
```diff
@@ -563,8 +563,7 @@ end
 Then you can get this error message in exception handler:
 ```ruby
 rescue_from Pundit::NotAuthorizedError do |e|
-  message = e.reason ? I18n.t("pundit.errors.#{e.reason}") : e.message
-  flash[:error] = message, scope: "pundit", default: :default
+  flash[:error] = e.reason ? I18n.t("errors.#{e.reason}", scope: "pundit") : e.message
```
astronaut-wannabe, line 566: since `"errors.#{nil}" == "errors."`, I think you can drop the ternary altogether 
```suggestion
  flash[:error] = I18n.t("errors.#{e.reason}", scope: "pundit", default: e.message)
```

astronaut-wannabe (approved):
I came across the same thing, lgtm!

dgmstuart:
We've decided to take a different approach. #684 reverts the change that this PR is based on.

PR #640, opened 2020-01-22
Update README.md

benkoshy:
add note re: protect_from_forgery

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Include Pundit in your application controller:
 ``` ruby
 class ApplicationController < ActionController::Base
   include Pundit
-  protect_from_forgery
+  protect_from_forgery # (Note: you don't need to explicitly add protect_from_forgery for rails 5.2 and above. It will be included in ActionController::Base)
 end
 ```
 


README.md
```diff
@@ -31,7 +31,7 @@ Include Pundit in your application controller:
 ``` ruby
   include Pundit
-  protect_from_forgery
+  protect_from_forgery # (Note: you don't need to explicitly add protect_from_forgery for rails 5.2 and above. It will be included in ActionController::Base)
```
dgmstuart: I think the point here was maybe to demonstrate that Pundit should happen first in the application controller?

Given that rails 4 is past end-of-life I guess we should just remove this line and add a comment saying that Pundit should come first in the controller (though I'm kind of guessing here - I don't know how important that actually is...)

adambutler:
I was just starting a new Rails 6 project and saw this in the README. I thought `protect_from_forgery` was a Rails method but wasn't 100% certain so searched this codebase for the method definition landing on this issue.

I think `protect_from_forgery` should be removed to remove confusion and instead change it to:

``` ruby
  include Pundit
   
   # ...
end
```

README.md
BKSpurgeon: I suspect it is out of convenience that pundit is added in the application controller - otherwise it may have to be added in where required. Shall I add in a note `protect_from_forgery` or remove it entirely?

dgmstuart:
This was resolved by https://github.com/varvet/pundit/pull/651

Issue #649, opened 2020-04-14
Doc: Is the README wrong, or am I having a massive brainfart?


rlue:
From the README (emphasis mine):

> Like with the authorize method, you can also override the policy scope class:
>
> ```ruby
> def index
>   # publication_class => Post
>   @publications = policy_scope(publication_class, policy_scope_class: PublicationPolicy::Scope)
> end
> ```
>
> Just as with your policy, this will **automatically infer that you want to use the `PostPolicy::Scope` class**, it will instantiate this class and call resolve on the instance. In this case it is a shortcut for doing:
>
> ```ruby
> def index
>   @posts = PostPolicy::Scope.new(current_user, Post).resolve
> end
> ```

What automatic inference is going on here? Based on my understanding of this section in context, it seems like it _should_ say:

> Just as with your policy, this will manually override the scope class used for this query. In this case it is a shortcut for doing:
>
> ```ruby
> def index
>   @publications = PublicationPolicy::Scope.new(current_user, Post).resolve
> end
> ```

Is that correct, or am I missing something?

Linuus:
Looks like you are correct 👍 

PR? :)

frenkel:
@Linuus this was fixed by the PR you merged earlier, so it can be closed :tada:

rlue:
I'll handle it :)

PR #650, opened 2020-04-14
Fix faulty explanation in README

rlue:
FYI: I opted to remove the entire sentence rather than amend it, as proposed in #649, because it didn't add anything new from the sentence before it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,13 +303,11 @@ def index
 end
 ```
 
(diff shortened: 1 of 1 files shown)


yagudaev:
This is great, should be merged in.

kajatiger:
Why is this not merged yet?

PR #653, opened 2020-05-09
Update README.md

natronic:
Make it clear that authorize controller code has to be placed at the top of controller method(s). Otherwise page will throw errors from any instance variables, etc. that get evaluated before authorization occurs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,10 @@ is what is passed as the record to `authorize` below.
 
 ```ruby
 # In controllers
-authorize :dashboard, :show?
+def show
+  authorize :dashboard, :show?
+  ...
+end
 ```
 
 ```erb


Linuus:
Can you give some more context here? There’s no requirement to call this method at any specific point from Pundits point of view.

yagudaev:
> Can you give some more context here? There’s no requirement to call this method at any specific point from Pundits point of view.

His point that is confusing to a rails new comer where to place the code. His suggestion shows it is to be called from a controller action. It is not a macro method and wouldn't make sense outside of the request/response cycle.

Burgestrand:
Thanks! We've written the documentation in a similar way for the other code examples, so this makes it consistent if nothing else 🎉

yagudaev:
Great job thanks Burgestrand, OSS is hard thanks for the hard work ❤️

Issue #654, opened 2020-05-14
Here we go again: Custom error messages?


rlue:
Sorry to beat a dead horse. People have been asking about custom error messages for over six years now (#66, #143, #212, #599). Clearly, there have been some improvements in the API since then (#114), and some great suggestions for how to implement this at the controller level (#503), but the approach currently recommended in the README still doesn't feel right to me.

### Official Approach

The README suggests one of two options:

* [Use i18n to generate error messages based on query, record, and policy](https://github.com/varvet/pundit#creating-custom-error-messages).
* [Raise errors inside your query methods](https://github.com/varvet/pundit#multiple-error-messages-per-one-policy-action).

#### What's wrong with this?

The first approach limits you to one error message per resource-query. Queries can fail for many reasons, and it's nice to be able to pass this information along in the exception message.

The second approach breaks the pattern of query methods returning a boolean. Pundit's own README recommends using them in views as follows:

```erb
<% if policy(@post).update? %>
  <%= link_to "Edit post", edit_post_path(@post) %>
<% end %>
```

Raising an error in `#update?` is not compatible with this approach.

### What I'm not asking for

A lot of people have asked to specify custom exceptions as an argument to `#authorize`. I am **100% against this idea**: if you're using policies in the first place, then that's the only class that should know about why authorization failed. If you spread that responsibility between the policy and the controller, then the policy just becomes a bucket for tossing helper methods into, which violates SRP and really defeats the purpose.

It appears that @jnicklas [shares this sentiment](https://github.com/varvet/pundit/pull/38#issuecomment-23259039).

### A proposal

#503 came very close to what I'd like to see, but it accomplishes custom error messages in `rescue_from`, via a combination of policies, controllers, and Rails' i18n module. As a library, I think Pundit should make minimal assumptions about external dependencies, and so I'd suggest something that lives entirely within Pundit itself—_i.e.,_ depending only on policy classes and `Pundit::NotAuthorizedError`:

```ruby
# lib/pundit.rb

class NotAuthorizedError < Error
  def initialize(options = {})
    ...
    message = options.[:message] || @policy.error_message || "not allowed to #{query} this #{record.inspect}" }
    ...
  end
end
```

#### Usage

```ruby
class ApplicationPolicy
  attr_accessor :error_message
end

class PostPolicy < ApplicationPolicy
  def update?
    self.error_message = if record.author != user
                           'Keep your grubby hands off other people’s stuff!'
                         elsif record.archived?
[...]

PR #655, opened 2020-05-14
Support custom error messages in policies

rlue:
This commit allows policies to define the error messages set on `Pundit::NotAuthorizedError` exceptions when `#authorize` fails. The rationale is described in detail in GitHub issue #654, and summarized below.

Some queries can fail in multiple ways; for instance,

```ruby
class PostPolicy
  def update?
    if record.author != user
      ... # failure case 1
    elsif record.archived
      ... # failure case 2
    end

    true
  end
end
```

In their controllers, users might wish to handle different failure modes in different ways, but prior to this commit, there was only one way to tell the difference—namely, by raising errors inside the query method:

```ruby
def update?
  if record.author != user
    raise Pundit::NotAuthorizedError, 'You’re not the author!'
  elsif record.archived
    raise Pundit::NotAuthorizedError, 'This post is old news!'
  end

  true
end
```

This breaks the expectation that query methods should return booleans, which in turn breaks a pattern for using query methods in views:

```erb
<% if policy(@post).update? %>
  <%= link_to "Edit post", edit_post_path(@post) %>
<% end %>
```

973b63b added a `reason` option to the NotAuthorizedError initializer, but ultimately required the same approach of raising errors in queries.

---

This commit enables a cleaner method of passing a custom error message to exceptions from within policies, without violating the expectations of where exceptions are raised from.

```ruby
class PostPolicy
  attr_accessor :error_message

  def update?
    self.error_message = if record.author != user
                           'You’re not the author!'
                         elsif record.archived
                           'This post is old news!'
                         end

    error_message.nil?
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,17 +483,24 @@ class NilClassPolicy < ApplicationPolicy
 end
 ```
 
-## Rescuing a denied Authorization in Rails
+## Handling `Pundit::NotAuthorizedError`
 
-Pundit raises a `Pundit::NotAuthorizedError` you can
-[rescue_from](http://guides.rubyonrails.org/action_controller_overview.html#rescue-from)
-in your `ApplicationController`. You can customize the `user_not_authorized`
-method in every controller.
+When `#authorize` fails, Pundit raises an error.
+For a coarse, one-line approach to handling all failed authorizations,
+set up an application-wide 403 Forbidden response
+any time this error is encountered:
+
+```ruby
+# application.rb
+
+config.action_dispatch.rescue_responses["Pundit::NotAuthorizedError"] = :forbidden
+```
+
+For more fine-grained control, [`rescue_from`][] this exception
+in your application controller:
 
 ```ruby
 class ApplicationController < ActionController::Base
-  include Pundit
-
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
   private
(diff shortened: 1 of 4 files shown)


take:
👍 for this

I think this approach is better than the current one included in master branch - https://github.com/varvet/pundit/commit/973b63b396c2a98099caf5eefd1c6841416eddfa

matteeyah:
Maybe the policy method could just return a failure reason or a custom query? Generating error messages should be handled at the presentation layer.

rlue:
> Generating error messages should be handled at the presentation layer.

Can you elaborate? Applications with no presentation layer at all (_e.g.,_ CLI apps) raise errors and set messages on them all the time; why should it be the sole province of the presentation layer in a web app?

The presentation layer can decide what is shown _to the user,_ and as the developer, you can decide whether that is `Pundit::NotAuthorizedError#message` or something else. IMO, setting a "failure reason" in the policy and then using that reason to generate a message in the controller is unnecessary indirection and complexity.

matteeyah:
I'm not saying it needs to be on the controller level, API-only applications have presentation layers as well. In a web application with a frontend it can be in a Presenter or a Decorator, in API applications it can be in the serializer and/or entity object.

Bubbling up messages along with the error is fine as well. As you said, some applications might not actually have a presentation layer. I just think we should be able to send a custom failure reason as well.

rlue:
> The library should be able to send a failure reason or an error message and leave the decision about how error messages are displayed to the application.

100% agree, but I also don't see why the error message displayed by the application has to be the same as the value of `Pundit::NotAuthorizedError#message`. Why not just set that to your "failure reason" and handle it accordingly in your presentation layer?

```ruby
def foo
  authorize(@post)
rescue Pundit::NotAuthorizedError => e
  case e.message
  when 'not author'
    # display one error message...
  when 'archived
    # display another error message...
  end
end
```

Maybe I am misunderstanding your original question:

> Maybe the policy method could just return a failure reason or a custom query symbol?

Can you provide an example of how this would be implemented, so I can see what you're talking aobut?

matteeyah:
Your example is almost exactly what I had in mind

The `NotAuthorizedError` also has `reason` and `query` fields that get bubbled up to the `rescue` block. (https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L25)

I figured the policy method could return a symbol that the error can be initialized with.

rlue:
Interesting! I assume you actually meant the following:

```ruby
def foo?
  return :no_user unless user
  return :archived if record.archived?

  true
end
```

because without `return`, the symbols are invoked in a null context and don't actually do anything.

The problem with this is that `#authorize` [expects the query method (`#foo?`) to return a boolean](https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L73):

```ruby
raise NotAuthorizedError, ... unless policy.public_send(query)
```

and symbols (like `:no_user` and `:archived`) are truthy. Thus, by returning a failure symbol, the query method is actually telling Pundit that authorization _succeeded._

This can be resolved by setting an attribute on the policy instead, which `#authorize` then uses to populate the error object:

```ruby
def foo?
  self.reason = :no_user unless user
  self.reason = :archived if record.archived?

  reason.nil?
end
```

which is basically what my original PR proposes, but with `#reason` instead of `#error_message`.

I don't know how the Pundit authors feel about adding even more new attributes to the policy class; personally, I think `#message` is flexible enough to do what you're asking, without violating any serious conceptual principles.

FunkyloverOne:
One way to deal with custom failures is right here, I use it a lot in my code. :)
https://dry-rb.org/gems/dry-monads/1.3/result/

Burgestrand:
Me and @dgmstuart had a chat about this, and I believe we both agreed that query methods should be encouraged to return a boolean (as opposed to raising an exception).

A possible problem with the approach suggested in this PR is that we've got memoized policy instances:

https://github.com/varvet/pundit/blob/2714875ac9b9f0de632ebaa886516d3e21c9724f/lib/pundit.rb#L256-L263

If I'm not wrong this means that the error reason here _could_ become very confusing in case you query the same policy multiple times, e.g. in a controller action:

```ruby
def do_something_in_a_controller
  if policy(record).can_do_this? # .error_message is now set
    # …
  elsif policy(record).can_do_that? # if this is true, .error_message is still going to be set from previous statement due to memoized instance
    # …
  else
     render policy(record).error_message
  end
end
```

I haven't verified this in actual code/test-case yet.

jeffse:
I'm running into a similar issue with wanting to provide a reason during authentication, but with the added feature of being able to provide a reason when authentication is _allowed_.

Here's my idea: create new classes `Pundit::Authorized` and `Pundit::NotAuthorized`.
```ruby
class Authorized
  attr_reader :reason

  def initialize(reason = nil)
    @reason = reason
  end
end

class NotAuthorized
  attr_reader :reason

  def initialize(reason = nil)
    @reason = reason
  end
end
```
Policies would be allowed to return one of these classes in addition to booleans.
```ruby
def create?
  return Pundit::NotAuthorized(:not_user) unless user
  return Pundit::Authorized(:admin) if user_has_admin_privileges?
  return Pundit::NotAuthorized(:not_editor) unless user.editor?

  true
end
```
Next, update the `authorize` method's logic as follows:
```ruby
result = policy.public_send(query)
if !result
  raise NotAuthorizedError, query: query, record: record, policy: policy
elsif result.is_a? NotAuthorized
  raise NotAuthorizedError, query: query, record: record, policy: policy, reason: result.reason
end
```
This has the advantage over the previously suggested approaches to not break existing code. It also gives some flexibility to the implementer as to whether they wanted to provide a symbol or a string for the `reason`. Personally, I'd be using symbols and updating my `pundit.en.yml` file to include the reason as a sub-category.

(`authorize` wouldn't handle the `Pundit::Authorized` case, but as I said at the beginning, I have a business need to be able to query this to also determine why a user has access to a particular action.)

Burgestrand:
Revisiting this today, and I don't think that having stateful policies as the default recommended approach is a good idea. Our main state keeping today is the policy/scope cache, and that's about it.

If it works for you, that's great, keep doing it! I'm mainly mindful of recommending it for all.

There's an approach posted in #654 that has a slightly different approach: https://github.com/varvet/pundit/issues/654#issuecomment-1564339565

I'd still love to hear how y'all end up solving this problem in your apps. Keep that in #654 though!

PR #660, opened 2020-08-25
Remove mention of old Rails versions from README

yagudaev:
No need to mention old version of rails, it takes away from the message. Plus with it being now Rails 6 and changes being introduced in Rails 4, we can assume users already have the strong params feature.


dgmstuart:
Seems reasonable 👍

PR #662, opened 2020-11-07
Add Truffleruby head to CI

gogainda:


gogainda:
@Linuus pls review/merge

dgmstuart:
@gogainda Hi - I'm not familiar with Truffleruby. If we introduce it to the build then we're to some extent committing to supporting it, and I'm not sure there's a good reason for us to do so? Let me know what you think.

gogainda:
Truffleruby is intended to be compatible with MRI ruby 2.7 . If you encounter any issues where Truffleruby build fails and MRI ruby 2.7 doesn't it means that it's up to Truffleruby team to take care of it in here https://github.com/oracle/truffleruby/issues

dgmstuart:
@gogainda the issue is that if a PR fails when building against Truffleruby, but passes for all other rubies, then that fails the build overall, and we're unlikely to be able to do anything about it ourselves.

My understanding is that the reason for building against multiple ruby versions is so that we can fix any issues in the Pundit source code if they happen to not be compatible with a particular ruby version. But it sounds like if we had a failure against Truffleruby, then it would be the Truffleruby team who would have to resolve it? I don't see why it's worthwhile for us to make this project dependent on that team.

@olleolleolle do you have a take on this?

olleolleolle:
@dgmstuart My simplest take: Add to the matrix! 

Implementations of Ruby which break could _if_ they cause issues that are not met with action from the language implementer/champion, be put in the `allow_failures` section.

dgmstuart:
Closed in favour of https://github.com/varvet/pundit/pull/678 which adds it to `allow_failures` - thanks @olleolleolle

PR #672, opened 2021-04-17
rescue-not_authorized_error-on-rspec

masarakki:
When apply [Multiple error messages per one policy action](https://github.com/varvet/pundit#multiple-error-messages-per-one-policy-action), 

`it { expect(subject).not_to permit(user, record) }` fails because it raise exception.
I think this matcher concerns only permit or not, so it should catch `Pundit::NotAuthorizedError`, and convert true/false.


dgmstuart:
@masarakki Hi - I think we've either changed or removed the suggestion on how to achieve "multiple" error messages: if I recall correctly it turned out to be quite a problematic approach. 

Could you please have a look through the current README and update the PR description accordingly.

PR #676, opened 2021-05-21
Build against currently supported rubies

dgmstuart:
2.6 is the earliest currently supported Ruby version:

  https://www.ruby-lang.org/en/downloads/branches/

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: 2.6
   Exclude:
     - "lib/generators/**/templates/**/*"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,22 @@
 language: ruby
+dist: focal
 
 matrix:
   include:
     - name: "RuboCop lint on pre-installed Ruby version"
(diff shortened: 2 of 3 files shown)


.rubocop.yml
```diff
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: 2.7
```
olleolleolle, line 2: Perhaps this should be 2.6, since RuboCop may offer changes that don't work in 2.6 with this setting?

dgmstuart: Sounds reasonable! Yes, I guess it should be the lowest supported version. 

Are you using JRuby at all? Am I correct in thinking 9.1 is old and not worth building against.

olleolleolle: Currently not using JRuby, and yes, 9.2 is the currently interesting version. 

Already from this, I think 9.1 should be let go from the CI matrix.

> JRuby 9.1.17.0 represents our stable release of our Ruby 2.3.x support

dgmstuart: Great - many thanks for the feedback 🙏

olleolleolle (approved):
I think this is ready to go, 🎉 !

PR #678, opened 2021-05-31
Build against Truffleruby head

dgmstuart:
Truffleruby is a high-performance implementation of Ruby:

  https://github.com/oracle/truffleruby

We don't want to be responsible for supporting it, so it goes in the
allow-failures section.

Based on https://github.com/varvet/pundit/pull/662

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,6 @@ matrix:
     - rvm: jruby-9.2.15.0
       env:
         - JRUBY_OPTS="--debug"
-
+    - rvm: truffleruby-head
+  allow_failures:
+    - rvm: truffleruby-head


olleolleolle:
👋 I think that the allow_failures section only annotates what _could_ match against a job.

So, the `include:` section needs the truffleruby line, too.

TIL: conditional there, huh. Unrelated to this.
https://docs.travis-ci.com/user/customizing-the-build/#conditionally-allowing-jobs-to-fail

dgmstuart:
@olleolleolle Oops- I thought I had added it to the main part of the matrix as well. Clearly not

olleolleolle (approved):
🎉

PR #680, opened 2021-07-28
Fix syntax issue in multiple error messages documentation

joshmcarthur:
`flash[key]` does not accept a scope and default after the equals sign as the README indicates. Instead, these keyword args should be provided to the call to I18n.t on the line above.

I think this was just a bad copy-paste when this documentation was originally added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,11 +566,11 @@ class ProjectPolicy < ApplicationPolicy
 end
 ```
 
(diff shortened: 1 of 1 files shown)


joshmcarthur:
Oops, just saw https://github.com/varvet/pundit/pull/638. Sorry for the noise!

PR #682, opened 2021-08-08
Policy spec template fix for RSpec/LeadingSubject

stgeneral:
Policy spec template fix for RSpec/LeadingSubject - https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/LeadingSubject.

### DIFF
--- a/lib/generators/rspec/templates/policy_spec.rb
+++ b/lib/generators/rspec/templates/policy_spec.rb
@@ -1,9 +1,9 @@
 require '<%= File.exists?('spec/rails_helper.rb') ? 'rails_helper' : 'spec_helper' %>'
 
 RSpec.describe <%= class_name %>Policy, type: :policy do
-  let(:user) { User.new }
-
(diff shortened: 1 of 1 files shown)


Burgestrand:
Hi! Thanks for the PR, but I won't merge it.

The reason for not merging this is the same to what I mentioned in your other PR: https://github.com/varvet/pundit/pull/683#issuecomment-902513187

In short I won't merge style changes that aren't enforced. Rubocop passes on `master` and because of that I can't see a logical reason to why it would make sense to merge this.

stgeneral:
Thank you for the detailed clarifications!

I guess that Rubocop passes on `master` because `RSpec/LeadingSubject` is a part of [rubocop-rspec](https://github.com/rubocop/rubocop-rspec/), which seems not included in `pundit.gemspec`.  I'll try to verify and confirm if that's the case.

PR #683, opened 2021-08-08
Policy spec template fix for Style/StringLiterals

stgeneral:
Update for the default setting of Style/StringLiterals - https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/StringLiterals.

### DIFF
--- a/lib/generators/rspec/templates/policy_spec.rb
+++ b/lib/generators/rspec/templates/policy_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe <%= class_name %>Policy, type: :policy do
 
   subject { described_class }
 
-  permissions ".scope" do
+  permissions '.scope' do
     pending "add some examples to (or delete) #{__FILE__}"
   end
 


Burgestrand:
Thank you very much, your PR is appreciated!

I'd like to merge this, I cannot do so in good conscience. The reason is that rubocop style is not enforced in the templates. It should be, but it is not. To merge this two things need to happen:

- [ ] Rubocop on Travis needs to run on our templates. Fixing that is desirable. An idea would be to generate the templates before running rubocop, and then running rubocop on the output which would cause fail/pass on Travis. I do not have the bandwidth to do this myself, so if you're willing to take a stab at fixing that and then requesting this PR to be reopened by all means please do.
- [ ] Change `.rubocop.yml` to accommodate the default style. Double quotes are explicitly configured in this repository. This point is the cause for many debates world-wide. I believe there's an argument to be made for following the default style. At the end of the day I honestly don't care which one we follow as long as it's enforceable. This does not need to happen in tandem with my first point, and probably shouldn't either.

Your underlying reasons for making this PR are well placed 🙌

Issue #656, opened 2020-05-23
New release?


iMacTia:
Latest release was August 2019, and since then there have been lots of commits in `master`.
Would the team consider releasing a new version with all the new fixes/features?

rlue:
Wait up! There are some commits in `master` that extend the API in a way that I and at least one other person [don't think are ideal](https://github.com/varvet/pundit/pull/655#issuecomment-632454600)—if a new release is coming, is there any chance we could get some feedback on #654 / #655 first? :pray:

iMacTia:
I understand the pending issues but the last release is almost 1 year old now and there hasn't been any movement on those issues since May/June 😕...

rlue:
If it helps, you can just [specify the commit you need in your Gemfile](https://bundler.io/guides/git.html#how-to-install-gems-from-git-repositories):

```ruby
gem 'pundit', git: 'https://github.com/varvet/pundit.git', ref: '504b86a'
```

iMacTia:
Of course! I did it back in May, but I hope you'll agree that's not a long-term solution 😅
I was hoping to see a proper release in a few weeks, definitely not expecting months!

vpereira:
I'm looking forward for a new release, specially because of https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L24-L41. Any plans for 2020?

mabras:
If anyone search for alternative:
https://actionpolicy.evilmartians.io/

gat-bryszard:
Any news in here? It doesn't look good with the last meaningful commit 17 months ago.

athix:
@kindell @olleolleolle @dgmstuart 

What are the long term plans for Pundit? I was planning on using Pundit as the primary example of Authorization combined with Sorcery as the Authentication, but it is a bit concerning if the release cadence is looking unhealthy.

Burgestrand:
Hi! I can't speak for the other maintainers, but I can give you some information from Varvet. This is all preliminary as we're still figuring things out at Varvet, and I can't run this post by @kindell or @dgmstuart since they're both (me too, really) on ~4-week swedish vacation and are off-grid right now. Please keep some spare oversight in hand while reading this.

That said, I just noticed this issue and I figured a vague reply is better than none.

**The short answer** is that we at Varvet have no intention of abandoning Pundit. It's likely that we'll continue spending resources on Pundit, and if that turns out to not be an alternative we are probably looking at finding a new maintainer.

[...]

PR #684, opened 2021-08-10
Revert "Merge pull request #625 from holyketzer/custom-messages"

Burgestrand:
Reverting this because it's blocking us from making a new release, see: https://github.com/varvet/pundit/issues/656#issuecomment-895827605

Additionally, me and @dgmstuart have backtracked a bit and believe that _in case_ you'd like to use a solution similar to this then you should probably use specialized error classes instead, something similar to this:

```ruby
UserProjectLimitReachedError = Class.new(Pundit::NotAuthorizedError)

def create?
  raise UserProjectLimitReachedError unless user.has_paid_subscription?
end
```

All that said, I believe that the reasoning in https://github.com/varvet/pundit/issues/654 is sound, and so query methods should strive towards returning truthy/falsy as opposed to raising errors to deal with custom error messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,46 +544,6 @@ en:
 Of course, this is just an example. Pundit is agnostic as to how you implement
 your error messaging.
 
-## Multiple error messages per one policy action
-
-If there are multiple reasons that authorization can be denied, you can show different messages by raising exceptions in your policy:
-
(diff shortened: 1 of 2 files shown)


dgmstuart (commented):
LGTM:+1:

PR #685, opened 2021-08-13
Update rubocop to most recent version

Burgestrand:
- Update Rubocop to most recent version.

I noticed #683 and #682 didn't fail the build even though they changed style, and our master branch hasn't failed either. If we have rubocop violations they should fail, and if we don't have violations in master I'd expect the PRs to fail since they don't change the configuration.

~~Through this I noticed that rubocop doesn't run for PRs. Hopefully this changes that.~~

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,5 @@
 AllCops:
+  NewCops: enable
   TargetRubyVersion: 2.6
   Exclude:
     - "lib/generators/**/templates/**/*"
@@ -15,7 +16,7 @@ Metrics/ModuleLength:
   Exclude:
     - "**/*_spec.rb"
 
-Metrics/LineLength:
+Layout/LineLength:
   Max: 120
 
 Metrics/AbcSize:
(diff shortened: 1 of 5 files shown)


Burgestrand:
Actually… we run the default rake task for all rubies. There's no reason to run rubocop multiple times, so it's better to run it separately.

dgmstuart:
@Burgestrand or... maybe I don't understand what's going on here: looking at the Travis build failure, rubocop doesn't seem to be running as part of the default rake task (which I think is a good thing), but is now correctly failing when we're explicitly running it in the Travis config.

I guess rubocop isn't running as part of the `rake` command because we have explicitly said that only `spec` should be run as the default task? https://github.com/varvet/pundit/blob/master/Rakefile#L20

I feel like I'm OK with the result, but I'm not sure if it's the result you intended?

Burgestrand:
Closing this because there's no clear benefit, it's causing more trouble than expected and there are other things to prioritise here. If you're reading this and feel updating rubocop and make the tests pass is something you'd like to do, by all means feel free to pick up where I left off!

Issue #689, opened 2021-09-23
Record is array when using namespaced policies with policy_class set


vivekmiyani:
## Ruby, Rails and pundit versions

Ruby version: 3.0
Rails version: 6.1.4.1
Pundit version: master

## How I am using

```ruby
# app/controllers/admin_controller.rb

class AdminController < ApplicationController
  def authorize(record, query = nil, policy_class: nil)
    super([:admin, record], query, policy_class: policy_class)
  end
end
```

```ruby
# app/controllers/admin/posts_controller.rb

class Admin::PostsController < AdminController
  def show
    @publication = authorize Post.find(params[:id]), policy_class: Admin::PublicationPolicy
  end
end
```

## Observed behaviour

Inside `Admin::PublicationPolicy` I am getting record array (with namespace).

```ruby
     4:       scope.all
     5:     end
     6:   end
     7:
     8:   def show?
 =>  9:     binding.irb
    10:   end
    11: end

3.0.2 :001 > record
 => [:admin, #<Post>]
```

## Expected behaviour

Inside `Admin::PublicationPolicy`, I should able to get `Post` object.

vivekmiyani:
Happy to raise an PR, if this is really an issue 😃 .

iggant:
Hi all. We have a trouble with current changes: 
we were using this feature as current manner: 

```ruby 
authorize([initial_country, target_country], policy_class: CountryPolicy)
```

and inside CountryPolicy: 

```ruby 
  def copy_to_country?
    initial_country, target_country = record

    country_allowed?(initial_country.country) && country_allowed?(target_country)
  end
```

and now there is no way to get initial_country. Record is always target_country

dgmstuart:
@iggant I'm afraid this doesn't look like a valid use of Pundit? 

When passing an array, this is interpreted as namespacing, as documented in the README: https://github.com/varvet/pundit#policy-namespacing

It seems that the thing which made your code work before was a bug in the code for handling namespacing, which was raised in this issue. 

Pundit is designed to authorise access to individual resources: passing arbitrary ruby objects (like this array of two objects) is not something we can reasonably support. 

I'm not sure I understand your use case without additional context, but in the code you've shown, there's no comparison between the two countries, so I believe it could be implemented like this: 

```ruby
authorize(initial_country.country)
authorize(target_country)
```

dgmstuart:
Or I guess maybe this?

```ruby
# controller
def copy 
  authorize(initial_country.country, :allowed)
[...]

PR #694, opened 2021-11-04
Pass record correctly when using `policy_class`

fqxp:
Fixes https://github.com/varvet/pundit/issues/689

This patch makes `authorize` remove namespaces before passing them to the constructor of `policy_class`, if defined.

Added tests accordingly.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -67,7 +67,7 @@ module Pundit
     # @raise [NotAuthorizedError] if the given query method returned false
     # @return [Object] Always returns the passed object record
     def authorize(user, record, query, policy_class: nil)
-      policy = policy_class ? policy_class.new(user, record) : policy!(user, record)
+      policy = policy_class ? policy_class.new(user, pundit_model(record)) : policy!(user, record)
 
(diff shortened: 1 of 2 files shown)


kevin-j-m:
Thanks for contributing this; I hope this gets resolved.

lib/pundit.rb
```diff
@@ -218,11 +218,12 @@ def authorize(record, query = nil, policy_class: nil)
 
     @_pundit_policy_authorized = true
 
-    policy = policy_class ? policy_class.new(pundit_user, record) : policy(record)
+    actual_record = record.is_a?(Array) ? record.last : record
+    policy = policy_class ? policy_class.new(pundit_user, actual_record) : policy(record)
 
     raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
```
toddkummer, line 224: This needs to have the same change, otherwise the error message reads "not allowed to update this Array"

```suggestion
    raise NotAuthorizedError, query: query, record: actual_record, policy: policy unless policy.public_send(query)
```

Burgestrand:
Thanks for this! We built on top of this in #697, and will be merging it shortly.

PR #697, opened 2022-01-04
Strip namespace when using `policy_class`

### LINKED ISSUE #666: authorize is returning array with namespace, not just record
> What did you do?

From a Rails controller action that includes Pundit, `project = authorize([:admin, Project.first])` as is done in the Pundit README.

> What did you expect to happen?

After Pundit uses `Admin::ProjectPolicy` to authorize the current user's access to `Project.first`, I expected `project` to be `Project.first`.

> What happened instead?

`project` was instead equal to `[:admin, Project.first]`.

> Make sure to include as much relevant information as possible. Ruby version, Pundit version, OS version and any stack traces you have are very valuable.

Ruby version 2.6.6
Pundit version 2.1.0
Mac OS 10.15.7

I see that `master` currently [handles the namespace array properly](https://github.com/varvet/pundit/blob/master/lib/pundit.rb#L226), but [the same method](https://github.com/varvet/pundit/blob/v2.1.0/lib/pundit.rb#L225) on version 2.1.0 just returns the array.

I validated that Pundit is finding the namespaced policy (`Admin::ProjectPolicy`) and using it correctly by throwing an exception when appropriate. This is simply an issue with the return value of `authorize` after a successful authorization using a namespaced policy.

Overall, I'm enjoying Pundit so far and would love to use this namespace shorthand, but until this is fixed, I'll need to explicitly pass the `policy_class` option or handle the returned array myself. Thanks for reading and in advance for any help on this!

### LINKED ISSUE #689: Record is array when using namespaced policies with policy_class set

Burgestrand:
Closes #689
Closes #694
Closes #666

Using `policy_class` and namespacing are two different ways of doing the same thing (specifying which policy to use). When both are given, we ignore the namespacing and rely on policy class since it's more specific — when we do so, we need to strip the namespace when passing the record to the policy class.

This is a refactor of #694.

TODO:
- [x] Update the changelog.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -67,7 +67,7 @@ module Pundit
     # @raise [NotAuthorizedError] if the given query method returned false
     # @return [Object] Always returns the passed object record
     def authorize(user, record, query, policy_class: nil)
-      policy = policy_class ? policy_class.new(user, record) : policy!(user, record)
+      policy = policy_class ? policy_class.new(user, pundit_model(record)) : policy!(user, record)
 
       raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
@@ -218,11 +218,12 @@ module Pundit
 
     @_pundit_policy_authorized = true
 
-    policy = policy_class ? policy_class.new(pundit_user, record) : policy(record)
+    actual_record = record.is_a?(Array) ? record.last : record
+    policy = policy_class ? policy_class.new(pundit_user, actual_record) : policy(record)
 
     raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
-    record.is_a?(Array) ? record.last : record
+    actual_record
   end
 
   # Allow this action not to perform authorization.
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -61,17 +61,23 @@ module Pundit
     # authorized to perform the given action.
     #
     # @param user [Object] the user that initiated the action
-    # @param record [Object] the object we're checking permissions of
+    # @param possibly_namespaced_record [Object, Array] the object we're checking permissions of
     # @param query [Symbol, String] the predicate method to check on the policy (e.g. `:show?`)
     # @param policy_class [Class] the policy class we want to force use of
+    # @param cache [#[], #[]=] a Hash-like object to cache the found policy instance in
     # @raise [NotAuthorizedError] if the given query method returned false
     # @return [Object] Always returns the passed object record
-    def authorize(user, record, query, policy_class: nil)
-      policy = policy_class ? policy_class.new(user, pundit_model(record)) : policy!(user, record)
+    def authorize(user, possibly_namespaced_record, query, policy_class: nil, cache: {})
+      record = pundit_model(possibly_namespaced_record)
+      policy = if policy_class
+        policy_class.new(user, record)
+      else
+        cache[possibly_namespaced_record] ||= policy!(user, possibly_namespaced_record)
+      end
 
       raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
 
-      record.is_a?(Array) ? record.last : record
+      record
     end
 
     # Retrieves the policy scope for the given record.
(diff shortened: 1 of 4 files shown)


sedubois:
How does one make sure that the namespace is preserved when *not* using `policy_class`?

With Pundit 2.1.1, `policy(content)` returned `ContentPolicy` and `policy([:admin, content])` returned `Admin::ContentPolicy`. With 2.2.0 it now seems to always return `ContentPolicy`. How to get the old behaviour back? NB: I'm looking for a generic solution working for any argument class (could be `Content`, `User`, etc).

sedubois:
I used to override `def policy(record)` in `Pundit::Authorization`, but that method now seems to be unused. `authorize` now calls `Pundit.authorize` which calls `Pundit.policy`. 

Should `def policy(record)` in `Pundit::Authorization` be deleted?

sedubois:
NB: I used to override `def policy(record)` (to add a namespace to the records and thus load namespaced policies when in an admin context) but that method is not called by `def authorize` any more, so it makes the overriding more difficult. Now I have to override both `def pundit` and `def authorize`.

dgmstuart:
@sedubois have you tried the suggestion laid out in https://github.com/varvet/pundit#policy-namespacing?

If it doesn't work, please raise an issue.

lib/pundit.rb
```diff
@@ -61,17 +61,23 @@ class << self
...
     # @raise [NotAuthorizedError] if the given query method returned false
     # @return [Object] Always returns the passed object record
-    def authorize(user, record, query, policy_class: nil)
-      policy = policy_class ? policy_class.new(user, record) : policy!(user, record)
+    def authorize(user, possibly_namespaced_record, query, policy_class: nil, cache: {})
+      record = pundit_model(possibly_namespaced_record)
+      policy = if policy_class
+        policy_class.new(user, record)
+      else
+        cache[possibly_namespaced_record] ||= policy!(user, possibly_namespaced_record)
```
adherr, line 75: This appears to be a breaking change, and I'm curious whether or not that was intentional. Our code uses the `authorize(record)` method below which [used to find the policy_class with `policy`](https://github.com/varvet/pundit/pull/697/files#diff-2acc48fe8302d7198f5a4aa4aaa8b4f0502d185e186a099283264da056d450c7L221) but this now uses `policy!` in that situation. 

Can you please verify that you meant to force a more explicit definition of `policy_class`? Thanks!

Burgestrand: Knee-jerk reaction is that's indeed unintended! We had some other things that were affected by this move that might also be unintended, so I'd wager it's very likely we'll make some adjustments to this code in the near future.

adherr: Thanks for the reply and opening a new issue. I'll stick to 2.1.1 until this is worked out. Let me know if there's anything I can do to help.

Burgestrand: Hi @adherr! This was a long time ago, but I'm here again for other reasons.

While the previous implementation used to use `policy(record)`, it still relied on the method returning an actual policy because the next call would be `policy.public_send(query)` which would fail with a `NoMethodError` if `nil` was returned for the policy.

The main change is that after this we instead raise `Pundit::NotDefinedError` as opposed to the `NoMethodError`.

What part of this behaviour were you relying on that prevented the upgrade?

Issue #701, opened 2022-01-07
Solve depreciations in rubocop.yml


MarceloAGuimaraes:
I was trying to solve some issues on Pundit, but my rubocop keeps failing because of wrong syntax and depreciations on rubocop config file.
Because of this, i can't go through some issues. This issue for example: https://github.com/varvet/pundit/issues/691.

This screenshot show some problems: 
<img width="347" alt="Screen Shot 2022-01-06 at 17 11 20" src="https://user-images.githubusercontent.com/41841441/148474122-2df35429-7936-458e-bf35-7aa0bfeff6da.png">

Burgestrand:
Hi! As I mentioned in https://github.com/varvet/pundit/pull/702#issuecomment-1007318138 there are no deprecations in `rubocop.yml` in the `main` branch. Are you perhaps running a newer version of rubocop on this repository instead of running `bundle exec rubocop`?

PR #702, opened 2022-01-07
updating rubocop config file

MarceloAGuimaraes:
https://github.com/varvet/pundit/issues/701

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,7 +15,7 @@ Metrics/ModuleLength:
   Exclude:
     - "**/*_spec.rb"
 
-Metrics/LineLength:
+Layout/LineLength:
   Max: 120
 
 Metrics/AbcSize:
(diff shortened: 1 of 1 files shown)


Burgestrand:
Hi! Thanks for doing this work, but your PR is _adding_ warnings to our rubocop CI, not removing them.

I'm trying to aggressively clean up this repo so I'll close this PR. That doesn't mean the discussion ends, but with the current rubocop version/config there are no `rubocop.yml` warnings in main.

MarceloAGuimaraes:
@Burgestrand , maybe we can think to upgrade the rubocop to a new version. They already released the 1.24, but the minimum ruby to this version is the 2.5.0. 
Do we have a problem with this requirement ?

Burgestrand:
> @Burgestrand , maybe we can think to upgrade the rubocop to a new version. They already released the 1.24, but the minimum ruby to this version is the 2.5.0. Do we have a problem with this requirement ?

That should be OK, the oldest version of Ruby that we support is typically the oldest non-EOL Ruby from https://www.ruby-lang.org/en/downloads/ — today that's 2.6. Anything older we can ignore.

MarceloAGuimaraes:
Awesome. I just created the pr https://github.com/varvet/pundit/pull/708.

PR #708, opened 2022-01-14
Update rubocop

MarceloAGuimaraes:


MarceloAGuimaraes:
I added a required_ruby_version to solve a log from rubocop, but i forgot about JRuby. I just removed the line from gemspec and disabled the rule(**Gemspec/RequiredRubyVersion**) that was requiring it.

Burgestrand:
> I added a required_ruby_version to solve a log from rubocop, but i forgot about JRuby. I just removed the line from gemspec and disabled the rule(**Gemspec/RequiredRubyVersion**) that was requiring it.

Nice work! Could you please update the commit message of https://github.com/varvet/pundit/pull/708/commits/08e23a1c25f3695c7339c7653e14249520a0c164 so that it explains _why_ it was removed?

An example would be:

> JRuby jruby-9.2 considers itself to be compliant with MRI 2.5, so adding a minimum required ruby version of 2.6 causes bundler to fail with the following error:
> 
> ```
> Resolving dependencies...
> Bundler found conflicting requirements for the Ruby version:
> 
>  In Gemfile:
>    Ruby (~> 2.5.8.0)
>
>    pundit was resolved to 2.1.1, which depends on
>      Ruby (>= 2.6.0)
> ```

MarceloAGuimaraes:
@Burgestrand makes totally sense. I just updated the commit message.

Burgestrand:
Awesome, thank you, well done!

PR #710, opened 2022-01-14
Add Ruby 3.1 to Travis

petergoldstein:
This PR adds Ruby 3.1.0 to CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ matrix:
         - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
     - rvm: 2.7.3
     - rvm: 3.0.1
+    - rvm: 3.1.0
     - rvm: jruby-9.2.17.0
       env:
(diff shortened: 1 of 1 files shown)


petergoldstein:
So I'm not sure why RVM doesn't have 3.1.0 available.  I've done a bunch of PRs adding Ruby 3.1.x to CI, but this is the first one I've tried to do with Travis - all the others have been GitHub Actions.  Anyone with some Travis experience know why the install is failing?

I use rbenv locally and have confirmed that bundle runs and rake runs green.

MarceloAGuimaraes:
Hey @petergoldstein, i don't have that much experience with Travis but i think the reason for this error is because the binary for ruby 3.1.x isn't available yet in **rvm** packages for ubuntu 20, as you can see here: https://rvm.io/binaries/ubuntu/20.04/x86_64/.

Burgestrand:
Hi! Travis fixed this about a week ago, and so this build passes when I re-ran the job. Thanks for the PR!

PR #712, opened 2022-01-15
Change Pundit::Authorization => Pundit

aurangzaib-danial:
`include Pundit::Authorization` raises uninitialised constant error. Therefore, use `include Pundit`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ Sponsored by:
 gem "pundit"
 ```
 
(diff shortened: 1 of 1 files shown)


Burgestrand:
Thank you. The documentation is not incorrect, the latest code is simply not yet released. If you're looking for documentation for Pundit you should be using [the documentation for the latest _released_ version](https://www.rubydoc.info/gems/pundit) as linked from the top of the README.

PR #715, opened 2022-01-19
Fix README with correct Controller include

haruska:
Update documentation for how to include Pundit in your application controller.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ Sponsored by:
 gem "pundit"
 ```
 
-Include `Pundit::Authorization` in your application controller:
+Include `Pundit` in your application controller:
 
 ``` ruby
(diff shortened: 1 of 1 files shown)


Burgestrand:
Hi! Thanks for the PR, but the README isn't wrong.

Please follow the relevant documentation for the version of Pundit you are using. You can find it at [rubydoc.info](https://www.rubydoc.info/find/gems?q=pundit). If you're using the latest released of version of Pundit then the most recent documentation is also linked to [rubydoc.info/gems/pundit](https://www.rubydoc.info/gems/pundit) at the top of the README.

The `main` branch on GitHub will sometimes be newer than the code you're actually using, which means that the code from the README on GitHub will not necessarily work in your application.

haruska:
Of course. My mistake.

Issue #696, opened 2021-12-16
Use of Struct in Headless Policies


drewmoore:
Thanks for this great gem! When trying to implement a headless policy, rubocop advised [against this](https://rubystyle.guide/#no-extend-struct-new). When experimenting on my own machine, I actually found it possible to avoid the usage of `Struct` altogether and instead just use the exact same pattern as for our ActiveRecord models' policies. This also works for me in views using `if policy(:my_resource).show?`. I'm wondering if Struct is needed, and if not, if the documentation could be updated to reflect this. Thanks!

Ruby version: 2.7.3
Pundit version: 2.1.1

dgmstuart:
Just came here to raise the same thing. 

There's also [a good blog post from The Pug Automatic](https://thepugautomatic.com/2013/08/struct-inheritance-is-overused/) going into this in some detail.

dgmstuart:
@drewmoore could you expand on what you mean by "instead just use the exact same pattern as for our ActiveRecord models' policies."?

dgmstuart:
@drewmoore We've discussed this and concluded that there isn't a particular reason to use Struct. We'd happily accept a PR which changes the documentation to a different approach if you have something specific in mind.

drewmoore:
> @drewmoore could you expand on what you mean by "instead just use the exact same pattern as for our ActiveRecord models' policies."?

Oh yes, I just meant that we are able to declare our headless policies in exactly the same manner as any other type of policy. I provided a documentation update on #717.

dgmstuart:
Closed by #171

PR #717, opened 2022-01-24
Remove Struct inheritance from headless policy example

drewmoore:
This addresses #696. Rubocop advises against inheriting from `Struct`, and my team found that we could simply remove this and achieve the same results. Thanks for your review.


README.md
```diff
@@ -194,7 +194,7 @@ you can retrieve it by passing a symbol.
 
 ```ruby
 # app/policies/dashboard_policy.rb
-class DashboardPolicy < Struct.new(:user, :dashboard)
+class DashboardPolicy
```
dgmstuart, line 197: I'm not sure this is enough: to me this suggests that just defining an empty class with a `Policy` suffix is enough, while in reality we need a very specific initialiser and accessors to be defined. 

How would you feel about explicitly writing out the initialiser and accessors as part of this documentation?

Presumably the motivation behind using `Struct` was as a shorthand for that (but I do see why it doesn't work).

Alternatively maybe it's worth making a `HeadlessPolicy` class to help avoid this repeated boilerplate 🤷

Burgestrand: Another option is to use the other `Struct` approach, which Rubocop is OK with?

```ruby
DashboardPolicy = Struct.new(:user, :dashboard) do
  # ...
end
```

drewmoore: True, I might also have come to that conclusion. Would this help to clarify? d17ca6c22279316e1cf65b58b351519f3797784d. As far as having an inherited class for headless policies, we actually just inherit from the same base policy class that all of our other policies use. The only difference is that the 2nd parameter of the initializer will just be the symbol passed to `authorize`, in this case `:dashboard`.

README.md
```diff
@@ -194,13 +194,22 @@ you can retrieve it by passing a symbol.
...
+
+  def show?
+    user.admin?
+  end
 end
 ```
 
 Note that the headless policy still needs to accept two arguments. The
-second argument will just be the symbol `:dashboard` in this case which
+first argument will just be the symbol `:dashboard` in this case which
```
Burgestrand, line 212: The `_record` is still `:dashboard` :)

drewmoore: I believe the first argument provided to `authorize` in this case would be `:dashboard`, whereas the second parameter in the policy's initializer method would in this case be `:dashboard`.

Burgestrand: Yup! The passage does talk about the headless policy, though?

dgmstuart: Is this an attempt to fix an unrelated issue? 

In this case the `authorize` method accepts only one argument, while the policy initializer accepts two. This line is pretty clearly talking about the policy initializer since the line before says: 

> Note that the headless policy still needs to accept two arguments. 

Maybe this area of the documentation could be clarified, but that feels completely independent of the change to remove Struct.

drewmoore: Changed: 67e0dd3417ee751ccbe1443b0c62e074e6ae2976

Burgestrand:
Thank you!

PR #719, opened 2022-01-26
Add note about GitHub documentation being more recent than what's released

Burgestrand:
We keep getting reported issues about the README being wrong, as people skip the table of contents.

Hopefully this note will help. I'd like to make it big and yellow, but I'm not sure how to call more attention to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ scalable authorization system.
 
 Links:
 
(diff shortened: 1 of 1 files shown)


Burgestrand:
This is easy to change and doesn't impact anything other than GitHub, so I'll be so bold and merge it even _before_ the review.

PR #724, opened 2022-03-08
Remove meaningless extra stripping of heredoc

anderscarling:
Squiggly heredoc is already stripped, so the `strip_heredoc` does not change the string.

Furthermore `strip_heredoc` is not available in early rails application setup, so we shouldn't use it here without `require 'active_support/core_ext/string/strip`

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -55,7 +55,7 @@ module Pundit
   class NotDefinedError < Error; end
 
   def self.included(base)
-    ActiveSupport::Deprecation.warn <<~WARNING.strip_heredoc
+    ActiveSupport::Deprecation.warn <<~WARNING
       'include Pundit' is deprecated. Please use 'include Pundit::Authorization' instead.
     WARNING
     base.include Authorization


dgmstuart (approved):
LGTM :+1::ship:

@Burgestrand this is a quick one - caused issues with Godmin on Rails 7 - what do you think?

anderscarling:
My best guesstimate is that this is due to pundit being loaded as a gem, thus being loaded/required quite early in the process of starting up the app.

anderscarling:
Looking at the backtrace I'm quite sure that's the issue:

```
lib/pundit.rb:58:in `included'
lib/godmin/authorization.rb:8:in `include'
lib/godmin/authorization.rb:8:in `<module:Authorization>'
lib/godmin/authorization.rb:5:in `<module:Godmin>'
lib/godmin/authorization.rb:4:in `<top (required)>'
lib/active_support/dependencies.rb:332:in `require'
lib/active_support/dependencies.rb:332:in `block in require'
lib/active_support/dependencies.rb:299:in `load_dependency'
lib/active_support/dependencies.rb:332:in `require'
lib/godmin.rb:10:in `<top (required)>'
test/dummy/config/application.rb:5:in `<top (required)>'          <------------- THIS IS WHERE GEMS ARE LOADED
                                                                                 Bundler.require(*Rails.groups)
```

`strip_heredoc` is most likely loaded [here in railties/lib/rails/application/bootstrap.rb](https://github.com/rails/rails/blob/main/railties/lib/rails/application/bootstrap.rb#L18) in an initialiser which would (i think) be loaded from `config/environment.rb` after loading `config/application.rb`.

Burgestrand:
Yeah. I think we can remove this call. Awesome!

PR #729, opened 2022-03-21
Use bundle add instead

glaucocustodio:
As per https://github.com/rubygems/rubygems/pull/5337, we can simplify the steps of adding a gem.


ioquatix (approved):

Burgestrand:
Looks good, `bundle add` is from at least 2017.

Burgestrand:
Thanks!

PR #737, opened 2022-05-19
Pass block as argument of describe call.

boof:
Hey there,

we recently running into this issue while trying to extend the permissions example group that the block cannot be extended as expected (eg. using module_eval in combination with instance_eval).

It also shortens the code. :)

### DIFF
--- a/lib/pundit/rspec.rb
+++ b/lib/pundit/rspec.rb
@@ -55,7 +55,7 @@ module Pundit
 
     module DSL
       def permissions(*list, &block)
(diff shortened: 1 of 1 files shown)


Burgestrand:
Hi! Sorry, you never received a reply.

What I'm missing here is an actual test of behaviour that this change fixes. I'm not against the change, but I don't understand _why_ the change.

Any code change carries some level of risk, and that needs to be balanced by some value received with the change usually demonstrated through an automated test.

PR #738, opened 2022-05-21
Add test for view "Helper" module to improve coverage

kenboo0426:
Hi, Thank you for the great Gem! 
I often use this gem!

Added `Helper.policy_scope` test to improve maintainability.
As a result, the test coverage changed from 98.41% to 100% in `lib/pundit.rb`

If there is a place to fix, please let me know.
best regards.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -216,6 +216,22 @@ class Controller
   end
 end
 
+class PostsView
+  include Pundit::Authorization
+  include Pundit::Helper
+  # Mark protected methods public so they may be called in test
+  # rubocop:disable Style/AccessModifierDeclarations
+  public(*Pundit::Authorization.private_instance_methods)
+  # rubocop:enable Style/AccessModifierDeclarations
+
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -216,6 +216,22 @@ def initialize(current_user, action_name, params)
   end
 end
 
+class PostsView
+  include Pundit::Authorization
+  include Pundit::Helper
+  # Mark protected methods public so they may be called in test
+  # rubocop:disable Style/AccessModifierDeclarations
+  public(*Pundit::Authorization.private_instance_methods)
+  # rubocop:enable Style/AccessModifierDeclarations
```
Burgestrand, line 225: Hi! Sorry about coming in this late.

I don't think we need to make any private instance methods public at all? Also the comment refers to `protected` methods but this does `private` methods.

What about this?

```suggestion
  def self.helper(mod)
    include(mod)
  end

  include Pundit::Authorization
```

PR #741, opened 2022-07-18
refactor: extract permitted_attribute_names from permitted_attributes

milgner:
In one of our controllers where we use delegated types and nested attributes, we needed to check the permitted attributes separately for the delegated type. This splits the method `permitted_attributes` in two, separating finding the attribute names and looking them up in the params, so that the attribute names can be retrieved individually.


Burgestrand:
Hi! Sorry about you not receiving a reply at all. 2022 was hectic, so while we did have a brief look the reply slipped our minds.

I'm generally positive about this. I'm slightly worried about expanding our public API, especially when the namespace is shared between pundit and whatever application is using this.

I'm pondering if we either prefix this method `pundit_permitted_attribute_names`, or if we expose this _only_ through the `Pundit::Context`. I need to simmer that for a bit.

PR #754, opened 2023-01-12
use File.exist? instead of File.exists?

johanandre:
File.exists? is deprecated in Ruby > 3.2. 
This causes an error when running the policy generator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ matrix:
     - rvm: 2.7.3
     - rvm: 3.0.1
     - rvm: 3.1.0
+    - rvm: 3.2.0
     - rvm: jruby-9.2.17.0
       env:
         - JRUBY_OPTS="--debug"
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
     - rvm: 3.0.1
     - rvm: 3.1.0
     - rvm: 3.2.0
-    - rvm: jruby-9.2.17.0
+    - rvm: jruby-9.3.10.0
       env:
         - JRUBY_OPTS="--debug"
     - rvm: truffleruby-head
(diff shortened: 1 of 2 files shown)


Obversity:
Can confirm that it's an issue for me too in a new-ish project, would be great to get this merged so people can use Ruby 3.2.

(Looks like the failing build is just Jruby and some incompatible ruby/bundler versions, I don't know enough about it to propose a fix sorry!)

dgmstuart:
@johanandre thanks for the PR. Looks like this is a breaking change for that version of JRuby - could you try updating the version of JRuby to 9.3? 

I can't work out if JRuby 9.2 is still supported (maybe @olleolleolle knows?). If so then either we need a solution which supports both usages, or to make a major version change.

olleolleolle:
@dgmstuart https://www.jruby.org/2022/06/27/jruby-9-2-21-0.html shows the last 9.2 release of JRuby.

That series implements Ruby 2.5. That Ruby is EOL.

I think it's fair not to keep supporting Ruby 2.5.

I also don't think it needs a major to drop support for an EOL Ruby version.

Hope this helps!

Burgestrand:
I relatively recently merged https://github.com/varvet/pundit/pull/758 which changes the lowest JRuby version to 9.3 — if you rebase on `main` this might be good to go!

johanandre:
@dgmstuart @olleolleolle I've updated the jruby version in the travis config to 9.3.10.0 which seems to be the oldest supported one. JRuby is kind of a mystery to me, so I hope that is correct!

Burgestrand (approved):
Thank you!

marcoroth:
Just wanted to check if there are any plans to release this fix in a patch release?

Burgestrand:
Hi @marcoroth this has now been released 😊

marcoroth:
Hey @Burgestrand, thank you for the release and the heads up! 🙏🏼

PR #758, opened 2023-02-16
Fix CI: Bump JRuby version to 9.3.x

harm-matthias-harms:
For recent PRs the pipeline fails because the JRuby version fails to install `bundler`. 

I recommend using JRuby 9.3.x as a test version because this is the matching version to the minimum tested ruby version 2.6.x.

> JRuby 9.3.10.0 is our point release of our Ruby 2.6.x support. Please check out our release notes for more information.
> https://www.jruby.org/download


Burgestrand (approved):
Great! Thank you!

PR #760, opened 2023-02-18
feat: Add customizable permit description

### LINKED ISSUE #759: Enable custom description for permit matcher
We're using RSpec output to document the permissions of all our resources. The default description is not what we want since it inspects the user and that adds noise to the output for no value.

I understand that it's probably not reasonable to change the default description because some people may depend on that behavior. But, it would be good to add the ability to customize it.

For example, I added a:

```ruby
module Pundit
  module RSpec
    module Matchers
      extend ::RSpec::Matchers::DSL
      def self.description(&block)
        @description = block.call
      end
      # ...
      matcher :permit do |user, record|
        # ...
        description do
          if Pundit::RSpec::Matchers.instance_variable_defined?(:@description)
            Pundit::RSpec::Matchers.instance_variable_get(:@description)
          else
            super
          end
        end
      # ...
end
```

If you're interested in a related PR, I'll send one in.

barelyknown:
In this common example, the default description is noisy and doesn't add any information:

🔈 Default output:

```
ActionItemPolicy
  show?
    grants access to a customer manager member
    when the user is a customer operations member
      when there is no relationship to the action item
        is expected not to permit #<User id: 2, email_address: "test+user2@x-b-e.com", mobile_number: nil, created_at: "2023-02-18 14:3...ation_event_at: nil, release_note_feed_token: nil, opt_out_of_check_in_request_notifications: false> and #<ActionItem id: 2, title: "Fix all the things!", responsible_organization_type: "Customer", responsi...quires_xbe_feature: false, meeting_id: nil, created_by_id: nil, project_id: nil, root_cause_id: nil>
```

This PR enables you to customize the description.

``` ruby
Pundit::RSpec::Matchers.description do
  "permit the user"
end
```

💆 Customized output:

```
ActionItemPolicy
  show?
    grants access to a customer manager member
    when the user is a customer operations member
      when there is no relationship to the action item
        is expected not to permit the user
```

- closes #759

### DIFF
--- a/README.md
+++ b/README.md
@@ -795,6 +795,14 @@ describe PostPolicy do
 end
 ```
 
+You can customize the description used for the `permit` matcher:
+
+``` ruby
+Pundit::RSpec::Matchers.description do
+  "permit the user"
+end
+```
+
 An alternative approach to Pundit policy specs is scoping them to a user context as outlined in this
 [excellent post](http://thunderboltlabs.com/blog/2013/03/27/testing-pundit-policies-with-rspec/) and implemented in the third party [pundit-matchers](https://github.com/chrisalley/pundit-matchers) gem.
 
--- a/lib/pundit/rspec.rb
+++ b/lib/pundit/rspec.rb
@@ -5,6 +5,10 @@ module Pundit
     module Matchers
       extend ::RSpec::Matchers::DSL
 
+      def self.description(&block)
+        @description = block.call
+      end
+
       # rubocop:disable Metrics/BlockLength
       matcher :permit do |user, record|
         match_proc = lambda do |policy|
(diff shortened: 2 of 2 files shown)


Burgestrand:
Thank you! I'm positive about this feature, but I'm not sure about the interface. I also need to borrow myself some time to test it out before I'm willing to merge it.

* I'm leaning to a regular module-level `attr_accessor` is more than good enough, similar (but not exactly) like what's in #761
```ruby
module Pundit
  module RSpec
    module Matchers
      class << self
        attr_accessor(:description)
      end
    end
  end
end
```

* _If_ we are to use a block, passing the `user` and `record` might be prudent to allow for this:

```ruby
Pundit::RSpec::Matchers.description = ->(user, record) { "permit user with role #{user.role}" }
```

bf4:
I merged my branch from https://github.com/varvet/pundit/pull/761 into this one

Burgestrand:
Hi! Merged in #806

PR #762, opened 2023-02-27
Opt-in for MFA requirement

tagliala:
Make the gem more secure by requiring that all privileged operations by any of the owners require OTP.

Ref: https://guides.rubygems.org/mfa-requirement-opt-in/


Burgestrand (approved):
Thank you! Turns out all gem owners of pundit are already using MFA, so I don't expect any trouble from this.

PR #764, opened 2023-04-11
Remove deprecated usage of ActiveSupport::Deprecation singleton

etiennebarrie:
Rails 7.1 will deprecate usage of the `ActiveSupport::Deprecation` singleton (calling `warn`, `silence`, etc. directly on it) (see https://github.com/rails/rails/pull/47354).

This remove the only use of it. I usually replace such usage by an instance of `ActiveSupport::Deprecation`, but in this case since it was only used once, and the caller information isn't as important as usual (it's pretty easy to find a call to `include Pundit`), I simply used a Ruby warning with `warn`. There's another difference in that the message isn't prepended by "DEPRECATION WARNING: ", let me know if that matters a lot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Use `Kernel.warn` instead of `ActiveSupport::Deprecation.warn` for deprecations (#764)
 - Policy generator now works on Ruby 3.2 (#754)
 
 ## 2.3.0 (2022-12-19)
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -55,7 +55,7 @@ module Pundit
   class NotDefinedError < Error; end
 
   def self.included(base)
-    ActiveSupport::Deprecation.warn <<~WARNING
+    warn <<~WARNING
       'include Pundit' is deprecated. Please use 'include Pundit::Authorization' instead.
     WARNING
     base.include Authorization
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -55,8 +55,10 @@ module Pundit
   class NotDefinedError < Error; end
 
   def self.included(base)
+    location = caller_locations(1, 1).first
     warn <<~WARNING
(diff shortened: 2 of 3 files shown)


Burgestrand:
Thank you! I think your assessment of things are correct.

I disagree with this:
> it's pretty easy to find a call to include Pundit

That's certainly true if the call is in your own code. However, when the `include Pundit` is from inside another gem it becomes much harder to find. It's not unlikely that a third-party Rails engine would use the deprecated method.

I'm not sure if there was an easy way to find the caller before, but I would really like it if we could make the offending code easier to find, even if it's just including `caller` in the deprecation warning message.

etiennebarrie:
> I'm not sure if there was an easy way to find the caller before, but I would really appreciate it if you could make the offending code easier to find; perhaps just by `caller` in the deprecation warning message.

Oh yes sorry that's something I was wrong about in the description, `ActiveSupport::Deprecation` adds a `DEPRECATION WARNING: ` prefix and a `(called from method at file:line)` suffix to deprecations. So previously it would look like this:

```
DEPRECATION WARNING: 'include Pundit' is deprecated. Please use 'include Pundit::Authorization' instead.
 (called from include at /Users/etienne/src/github.com/etiennebarrie/rails-app/app/controllers/application_controller.rb:3)
```

I copied it, including the space at the beginning of the second line.

Also fixed the output from `warn` appearing when running the other spec related to that warning.

Burgestrand (dismissed):
This is great, thank you!

I should have asked about this sooner. Could you please update the `CHANGELOG.md` with a note about this change, too?

etiennebarrie:
I rebased to deal with conflicts. I guess I should squash this all? Unless you squash on merge?

Burgestrand:
> I rebased to deal with conflicts. I guess I should squash this all? Unless you squash on merge?

Great! We don't squash here unless there's a strong reason to. We prefer to keep individual commits meaningful, and included. I believe your commits here are worthy of being allowed to live forever 😁

Burgestrand (approved):
LGTM ❤️

PR #765, opened 2023-04-28
Add the keyword args patch to pass args to the permission methods.

craigmcnamara:
I don't think this is a good pattern, but making it work is easier than rewriting all the bad code in our app. Going to add deprecations sometime soon so we can clean it up in a controlled manner.


craigmcnamara:
Whoops, this wasn't supposed to be sent to the main project.

Burgestrand:
No worries, it happens 😊

PR #763, opened 2023-04-11
Fixed travis branch

akmhmgc:

PR #766, opened 2023-05-04
Fixed travis branch

Burgestrand:
This is from #763, but the URL needed to be adjusted to Travis' new format.


Burgestrand:
Merging. Build is passing, but github doesn't get the new status.

Issue #723, opened 2022-02-20
Overriding both authorize and policy is required for namespacing


sedubois:
The [README](https://github.com/varvet/pundit#policy-namespacing) explains how to namespace policies, however this does not take into account direct calls to `policy`. Before the merge of #697, it was possible to only override `policy` and `authorize` would automatically be properly scoped as well. Now it has become necessary to override both `authorize` and `policy` in `Pundit::Authorization`. It still works, but is just now more cumbersome, and also required some figuring out when performing the upgrade to Pundit 2.2.0.

Opening this issue [as suggested in #697](https://github.com/varvet/pundit/pull/697#issuecomment-1046029897).

dgmstuart:
I'm inclined to say that the fact that `authorize` was implemented using `policy` was an implementation detail, and that it's not something which should be relied on.

However, you're absolutely right that there's nothing written in the documentation about how to namespace direct calls to `policy` and how to override calls to `policy` in the `AdminController` example. Before doing that, we'd need to add some tests for namespacing those calls - it's not something we have time to do right now, but it's something we'd like to see.

igorlvicente:
I want to tackle this issue but I suspect I have not quite understood what it is to be done.
@dgmstuart, can you explain better what you have in mind?
@sedubois, can you tell me whether what @dgmstuart is suggesting will solve the problem you - or others - have?

mattzollinhofer:
> I'm inclined to say that the fact that `authorize` was implemented using `policy` was an implementation detail, and that it's not something which should be relied on.

I can understand the perspective the `policy` was an implementation detail, but it was the implicit link between a custom policy's use for `authorize` calls and view's usage of custom policy so it's a little bit of a difficult balance for me to understand.

I opened a new issue (#740) trying to more clearly call this detail out as it's not necessarily related to namespacing in my case.

Burgestrand:
Now with `Pundit::Context` I believe we could provide namespacing by overriding the `pundit` context in the controller: https://github.com/varvet/pundit/blob/7e59b98a72850a02c0a10ec17fff425cff18d31a/lib/pundit/authorization.rb#L19-L24

There's no explicit implementation for that, but I'm thinking something like:
```
def pundit = super.with_namespace(...)
```

PR #772, opened 2023-05-23
Adds tests to cover policy overrides and when authorize is called in other namespaces

TheDevJoao:
This PR adds more tests as suggested in [this](https://github.com/varvet/pundit/issues/723#issuecomment-1085617779) comment from the issue [#723 ](https://github.com/varvet/pundit/issues/723)

## To do

- [X] I have read the [contributing guidelines](https://github.com/varvet/pundit/contribute).
- [X] I have added relevant tests.
- [X] No documentation alterations required.
- [X] I have made sure the individual commits are meaningful.
- [] I have added relevant lines to the CHANGELOG.

### DIFF
--- a/spec/authorization_spec.rb
+++ b/spec/authorization_spec.rb
@@ -8,6 +8,7 @@ describe Pundit::Authorization do
   let(:post) { Post.new(user) }
   let(:customer_post) { Customer::Post.new(user) }
   let(:comment) { Comment.new }
+  let(:admin_post) { Project::Admin::Post.new(user) }
   let(:article) { Article.new }
   let(:article_tag) { ArticleTag.new }
   let(:wiki) { Wiki.new }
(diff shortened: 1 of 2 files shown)


TheDevJoao:
Hey @dgmstuart, could you check this out in the near future? I'd appreciate any feedback, thanks! 😄

spec/authorization_spec.rb
```diff
@@ -116,6 +117,39 @@
     it "raises an error with a invalid policy constructor" do
       expect { controller.authorize(wiki, :destroy?) }.to raise_error(Pundit::InvalidConstructorError)
     end
+
+    context "when passed a record with the admin namespace" do
+      it "returns the record when the permission check passes" do
+        allow(controller).to receive(:authorize).with(admin_post, :update?).and_return(admin_post)
+        expect(controller.authorize(admin_post, :update?)).to eq(admin_post)
+      end
```
dgmstuart, line 125: This test doesn't seem to be testing the application code at all? 

`allow(controller).to receive(:authorize)` creates a stub , so `allow(controller).to receive(:authorize)` is only exercising that stub, not actually calling the `authorize` method on the controller code.

dgmstuart (changes requested):
Thanks for the submission. 

A number of these tests seem to be stubbing (eg. `allow(controller).to receive(x)...`) and then testing the _stub_ (`expect(controller.x)....`). 

It seems like these tests aren't actually exercising the application code. An important principle to follow is [never stub the system under test](https://thoughtbot.com/blog/don-t-stub-the-system-under-test).

Please remove those tests or change them so that they're actually exercising the Pundit code.

spec/authorization_spec.rb
TheDevJoao: I appreciate the feedback, @dgmstuart! 

The blog post you linked was a great read, I'll refactor these tests asap.

TheDevJoao:
@dgmstuart Any updates on this PR?

spec/authorization_spec.rb
```diff
@@ -159,6 +168,38 @@
...
+    end
+
+    context "when the policy is overriden" do
+      it "returns the correct overriden policy for a Post record" do
+        post = Post.new(user)
+
+        policy = controller.policy(post)
+        expect(policy).to be_instance_of(PostPolicy)
+        expect(policy.user).to eq(user)
+        expect(policy.post).to eq(post)
```
dgmstuart, line 185: This `context` says "when the policy is overridden", but it doesn't look like anything is being overriden? What's the "policy" which is being overridden (do you mean the `policy` method, or the `Policy` class?), and where is it overridden? 

I see there are some new things added in the `spec_helper` but none of those override `policy_class`, `authorize`, `policy`, `Post` or `PostPolicy`, so this spec looks like it just tests that `::PostPolicy` works - nothing to do with any overriding? 

What am I missing? 

If I understand it correctly, the "overriding" which the original issue was talking about defining `policy` or `authorize` in a specific controller - eg. in an admin controller you might want to do something like this:

```ruby
    super([:admin, record], query)
  end

  def policy(record)
    super([:admin, record])
  end 
```

...to make sure that calls to `authorize(Post)` and `policy(Post)` in that controller end using the `Admin::PostPolicy` instead of `::PostPolicy`.

I'm actually using this exact override of `authorize` in my current project.

dgmstuart (changes requested):
Hi - so sorry for the absurd delay - I've been crunching on a project for a long stretch and haven't had any headspace for Pundit. 

I'm afraid I don't understand what these specs are intended to cover: I need a little more explanation.

PR #776, opened 2023-05-26
Do not use NotImplementedError

### LINKED ISSUE #775: Do not use NotImplementedError
This error is not appropriate for method not implemented.

https://ruby-doc.org/3.2.2/NotImplementedError.html

> Raised when a feature is not implemented on the current platform. For example, methods depending on the `fsync` or `fork` system calls may raise this exception if the underlying operating system or Ruby runtime does not support them.
> 
> Note that if `fork` raises a `NotImplementedError`, then `respond_to?(:fork)` returns `false`.

This error is not rescued by this syntax as its [ancestor is not `StandardError` but `ScriptError` ](https://oleg0potapov.medium.com/ruby-notimplementederror-dont-use-it-dff1fd7228e5)

```ruby
class Foo
  def self.bar
    raise NotImplementedError, "You must define #resolve in #{self.class}"
  rescue => e
    puts e.message
  end
end

# IRB

irb(main):015:0> Foo.bar
(irb):10:in `bar': You must define #resolve in Class (NotImplementedError)
	from (irb):15:in `<main>'
	from /Users/hery/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.4/exe/irb:9:in `<top (required)>'
	from /Users/hery/.asdf/installs/ruby/3.2.2/bin/irb:25:in `load'
	from /Users/hery/.asdf/installs/ruby/3.2.2/bin/irb:25:in `<main>'
```

hallelujah:
Instead use `NoMethodError` as to indicate that the method should be defined in the class

Closes #775

## To do

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-Nothing.
+- Policy generator uses `NoMethodError` to indicate `#resolve` is not implemented (#776)
 
 ## 2.3.1 (2023-07-17)
 
--- a/lib/generators/pundit/install/templates/application_policy.rb
+++ b/lib/generators/pundit/install/templates/application_policy.rb
@@ -43,7 +43,7 @@ class ApplicationPolicy
     end
 
     def resolve
-      raise NotImplementedError, "You must define #resolve in #{self.class}"
+      raise NoMethodError, "You must define #resolve in #{self.class}"
     end
 
     private
--- a/spec/generators_spec.rb
+++ b/spec/generators_spec.rb
@@ -35,7 +35,7 @@ RSpec.describe "generators" do
       describe "#resolve" do
         it "raises a descriptive error" do
           scope = WidgetPolicy::Scope.new(double("User"), double("User.all"))
-          expect { scope.resolve }.to raise_error(NotImplementedError, /WidgetPolicy::Scope/)
+          expect { scope.resolve }.to raise_error(NoMethodError, /WidgetPolicy::Scope/)
         end
       end
     end


denisdefreyne:
I worry that `NoMethodError` will cause confusion, because the method *does* exist; it just isn’t properly implemented in subclasses.

In my own code, I’ve used my own `SubclassResponsibilityError`, like this:

```ruby
class SubclassResponsibilityError < StandardError
  def initialize(klass, method_name)
    super("Subclasses of #{klass} must implement the ##{method_name} method.")
  end
end
```

Would this make sense? It would mean adding a custom exception to the generator, which might not be what we want.

In any case, I do think `NoMethodError` is an improvement over `NotImplementedError`, so this PR gets a thumbs up from me!

Burgestrand:
@denisdefreyne I understand your worry, and it's a valid point. 😊

Taking that into consideration, and weighing that against adding a custom exception to the generator, together with the fact that the error message is descriptive, I think it's worth seeing how this flies in the wild.

It's generated code, so if this turns out to be confusing for people then 1) we can adjust it and 2) nobody needs to wait for a new release.

Linuus:
What is a bit odd about this to me is that `respond_to?(:resolve) #=> true` but then raise `NoMethodError` when called.

https://ruby-doc.org/core-3.0.0/NoMethodError.html
> Raised when a method is called on a receiver which doesn't have it defined and also fails to respond with method_missing.

I'm not saying `NotImplementedError` is perfect:
https://ruby-doc.org/core-3.0.0/NotImplementedError.html
> Raised when a feature is not implemented on the current platform. For example, methods depending on the fsync or fork system calls may raise this exception if the underlying operating system or Ruby runtime does not support them.

But maybe it maps closer to how it's used in this context?

Burgestrand (approved):
Thank you!

My thinking is that this error is _for developers_ and not something you would expect to see runtime. We could have our own error (`MethodNotImplemented = Class.new(Pundit::Error)`) and raise that, but I don't think increasing our API surface area (another public name) is worth it for this.

Main benefit here is no longer inheriting from `ScriptError`, which an unimplemented `resolve` definitely is not.

Burgestrand:
FYI @hallelujah I updated this branch with a rebase, because `main` ran into a build issue on jruby due to no fault of this PR.

franzliedke:
Strictly speaking, this is a BR break. `NotImplementedError` is not a subclass of `StandardError`, whereas `NoMethodError` is. That means the exception will now be handled by a `rescue` statement (without specific exception class).

Burgestrand:
This change affects a generator for the ApplicationPolicy, which is typically run once to get you started. You should be able to upgrade Pundit just fine, without this PR breaking anything.

franzliedke:
Of course, totally missed that it's in a generator. 🤦🏼

PR #787, opened 2023-10-13
Refactor pundit policy and policy_scope

bhtabor:
Refactor pundit policy and policy_scope

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -99,13 +99,7 @@ module Pundit
       policy_scope_class = PolicyFinder.new(scope).scope
       return unless policy_scope_class
 
-      begin
-        policy_scope = policy_scope_class.new(user, pundit_model(scope))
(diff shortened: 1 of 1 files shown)


Burgestrand:
Hi! Thank you for the PR.

For me, refactoring needs to have a goal. Either by making something easier to implement, or something easier to understand.

This refactor:
* Doesn't implement anything, so the refactor is not in order to "make the change easy".
* It adds indirection, which makes code _harder_ to understand.

It does indeed reduce duplication which _could_ be a good thing, but _in this PR_ it's not apparent what that might be.

bhtabor:
> For me, refactoring needs to have a goal. Either by making something easier to implement, or something easier to understand.

Hi @Burgestrand, thank you for reviewing. The reason I chose to send a refactor PR was because the duplicate error handling in the bang methods made it seem like error handling was different for the bang methods.

PR #789, opened 2023-11-01
Add policy_class to permitted_attributes

### LINKED ISSUE #742: Add `policy_class` parameter to `permitted_attributes` function
Is it possible to add `policy_class` parameter to `permitted_attributes` function (similar to `authorise`)?

So I can call:

```
permitted_attributes(@post, :update, policy_class: MyPolicy)
```

This is necessary for my code where the name of the policy and record class are different.

hobbypunk90:
Fixes #742 again 😅

Sorry, i cleaned my github a bit too fast.

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -80,7 +80,8 @@ module Pundit
       policy = if policy_class
         policy_class.new(user, record)
       else
-        cache[possibly_namespaced_record] ||= policy!(user, possibly_namespaced_record)
+        cache[{ policy_class: policy_class,
+                record: possibly_namespaced_record }] ||= policy!(user, possibly_namespaced_record)
       end
 
       raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
--- a/lib/pundit/authorization.rb
+++ b/lib/pundit/authorization.rb
@@ -98,9 +98,14 @@ module Pundit
     #
     # @see https://github.com/varvet/pundit#policies
     # @param record [Object] the object we're retrieving the policy for
+    # @param policy_class [Class] the policy class we want to force use of
     # @return [Object, nil] instance of policy class with query methods
-    def policy(record)
-      policies[record] ||= Pundit.policy!(pundit_user, record)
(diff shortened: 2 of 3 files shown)


lib/pundit.rb
```diff
@@ -80,7 +80,8 @@ def authorize(user, possibly_namespaced_record, query, policy_class: nil, cache:
       policy = if policy_class
         policy_class.new(user, record)
       else
-        cache[possibly_namespaced_record] ||= policy!(user, possibly_namespaced_record)
+        cache[{ policy_class: policy_class,
```
Burgestrand, line 83: This is always `nil` here.

Burgestrand:
Hi! Thanks for the PR!

I posted a comment in the ol' issue https://github.com/varvet/pundit/issues/742#issuecomment-2026343020 about what I think we should probably do.

I'd rather avoid having to modify other methods when adding support for this, to avoid breaking existing code.

By adding a `policy_class: nil` parameter to `permitted_attributes` and _only_ altering behaviour if that is passed we can probably add this feature without risking altering behaviour of existing code.

lib/pundit/authorization.rb
```diff
@@ -98,9 +98,14 @@ def policy_scope(scope, policy_scope_class: nil)
     #
     # @see https://github.com/varvet/pundit#policies
     # @param record [Object] the object we're retrieving the policy for
+    # @param policy_class [Class] the policy class we want to force use of
     # @return [Object, nil] instance of policy class with query methods
-    def policy(record)
-      policies[record] ||= Pundit.policy!(pundit_user, record)
+    def policy(record, policy_class: nil)
+      policies[{ policy_class: policy_class, record: record }] ||= if policy_class
```
Burgestrand, line 104: While `policies` is technically marked as private, I'm cautious against modifying its behaviour/content.

We probably should cache when `policy_class` is passed somehow, but that's a bigger consideration than adding the feature to `permitted_attributes`.

Burgestrand (changes requested):
I posted some comments :)

PR #792, opened 2023-12-24
Qualify the `Scope` class name

andyw8:
I'd like to propose this small change to the generator, and the corresponding examples in the README. However, I'm very new to Pundit so I'm seeking input from others with more experience.

When using the generator, it results in a `Scope < Scope` inheritance, equivalent to `Scope < ApplicationPolicy::Scope`.

This may be a matter of personal taste, but I find that having a class inheriting from another with (visually) the same name is jarring, and reduces readability, since it requires reading the surrounding context.

Additionally, this causes complications with typing system. For example, if using Sorbet:

```
app/policies/list_policy.rb:37: Circular dependency: ListPolicy::Scope is a parent of itself https://srb.help/5011
    37 |  class Scope < Scope
```
(although it can be argued that this is a bug/limitation of Sorbet).

Looking forward to hearing some feedback.

### DIFF
--- a/lib/generators/pundit/policy/templates/policy.rb
+++ b/lib/generators/pundit/policy/templates/policy.rb
@@ -1,6 +1,12 @@
 <% module_namespacing do -%>
 class <%= class_name %>Policy < ApplicationPolicy
-  class Scope < Scope
+  # NOTE: Up to Pundit v2.3.1, the inheritance was declared as
+  # `Scope < Scope` rather than `Scope < ApplicationPolicy::Scope`.
+  # In most cases the behavior will be identical, but if updating existing
+  # code, beware of possible changes to the ancestors:
+  # https://gist.github.com/Burgestrand/4b4bc22f31c8a95c425fc0e30d7ef1f5
+
+  class Scope < ApplicationPolicy::Scope
     # NOTE: Be explicit about which records you allow access to!
     # def resolve
     #   scope.all
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Update `ApplicationPolicy`` generator to qualify the `Scope` class name (#792)
+
 - Policy generator uses `NoMethodError` to indicate `#resolve` is not implemented (#776)
 
 ## 2.3.1 (2023-07-17)
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ generator, or create your own base class to inherit from:
 
 ``` ruby
 class PostPolicy < ApplicationPolicy
-  class Scope < Scope
+  class Scope < ApplicationPolicy::Scope
     def resolve
       if user.admin?
         scope.all
@@ -476,7 +476,7 @@ example, associations which might be `nil`.
 
 ```ruby
 class NilClassPolicy < ApplicationPolicy
-  class Scope < Scope
+  class Scope < ApplicationPolicy::Scope
     def resolve
       raise Pundit::NotDefinedError, "Cannot scope NilClass"
     end


Burgestrand:
> This may be a matter of personal taste, but I find that having a class inheriting from another with (visually) the same name is jarring, and reduces readability, since it requires reading the surrounding context.

I agree. I think your proposal is great. We should merge it.

However.

This changes behaviour compared to code generated before this change in a subtle way: https://gist.github.com/Burgestrand/4b4bc22f31c8a95c425fc0e30d7ef1f5

I don't think code should take advantage of `< Scope` being different from `< ApplicationPolicy::Scope`, but even more strongly I believe Pundit should avoid unpleasant surprises between updates.

I'm uncertain if a mere CHANGELOG entry is sufficient, and I think a comment in the generated code would be good. Generated code should always be inspected, so a comment should be sufficient to highlight this change to avoid developers being caught off-guard.

Could you please:
* Update the CHANGELOG.
* Add a comment to the policy generator template to highlight that the code might be different from previously generated policies.

Burgestrand (changes requested):
> This may be a matter of personal taste, but I find that having a class inheriting from another with (visually) the same name is jarring, and reduces readability, since it requires reading the surrounding context.

Thank you for the PR! I agree. I think your proposal is great. We should merge it.

However.

This changes behaviour compared to code generated before this change in a subtle way: https://gist.github.com/Burgestrand/4b4bc22f31c8a95c425fc0e30d7ef1f5

I don't think code should take advantage of `< Scope` being different from `< ApplicationPolicy::Scope`, but even more strongly I believe Pundit should avoid unpleasant surprises between updates.

I'm uncertain if a mere CHANGELOG entry is sufficient, and I think a comment in the generated code would be good. Generated code should always be inspected, so a comment should be sufficient to highlight this change to avoid developers being caught off-guard.

Could you please:
* Update the CHANGELOG.
* Add a comment to the policy generator template to highlight that the code might be different from previously generated policies.

lib/generators/pundit/policy/templates/policy.rb
```diff
@@ -1,6 +1,12 @@
 <% module_namespacing do -%>
 class <%= class_name %>Policy < ApplicationPolicy
-  class Scope < Scope
+  # NOTE: Prior to Pundit v2.4.0, the inheritance was declared as
```
andyw8, line 3: (I'm assuming v2.4.0 will be the next release?)

lib/generators/pundit/policy/templates/policy.rb
```diff
@@ -1,6 +1,12 @@
...
+  # `Scope < Scope` rather than `Scope < ApplicationPolicy`. In most cases the
+  # behavior will be identical, but if updating exisiting code, beware of
+  # possible changes to the ancestors:
+  # https://gist.github.com/Burgestrand/4b4bc22f31c8a95c425fc0e30d7ef1f5
```
andyw8, line 7: Rather than adding a long explanation as a comment, I thought it better to link to your illustration. If you'd prefer not to have a link to your gist, then maybe there could be a page in the repo?

andyw8:
@Burgestrand thanks for the feedback! I've pushed some updates.

lib/generators/pundit/policy/templates/policy.rb
Burgestrand: It depends! Right now there are no strict improvements or new features, so I'd probably make it `2.3.2` if released today. Even "prior releases of pundit" should be sufficient, as hopefully that's enough to make people take an extra look.

lib/generators/pundit/policy/templates/policy.rb
Burgestrand: I think that's fine. I'll probably forget about the gist and it'll be there for however long github lets it stay there. The gist also has a comment pointing back to this PR, so if I ever end up in a cleaning mood I'll probably see this and remember to let it be 🙂

Burgestrand:
I think this is good! I'll give you a chance to read the comments, and if you're up for it removing or adjusting the version specificer in the comment. Once you're happy we'll merge!

lib/generators/pundit/policy/templates/policy.rb
Burgestrand: Could also specify "up to 2.3.1" :D

andyw8:
Thanks, updated.

Burgestrand (approved):

Burgestrand (approved):

PR #793, opened 2024-01-02
Fix CodeClimate badge

tagliala:
Use snippet from https://codeclimate.com/github/varvet/pundit/badges#maintainability-markdown

[ci skip]


Burgestrand (approved):
Thank you!

Issue #735, opened 2022-05-09
[Request] Generate policy file when using scaffolding


james-em:
Hi,

It would be great if running

```
rails g scaffold Model
rails g controller Model
```

would actually generate an empty policy file that inherits from ApplicationPolicy.

As a bonus, for resources scaffolding, maybe we could implements all CRUD actions and defaults to false.

james-em:
Linking this in case it helps the implementation https://gist.github.com/annikoff/331f785aa7a207a7945b1eca6eff526b

Burgestrand:
FYI this came up again with #785, so I suggested moving that into here.

hoverjet:
Hi, I've added https://github.com/varvet/pundit/pull/794

55728:
Hi @Burgestrand! I'd like to take a shot at this.

Based on your feedback in #794, I understand that:
- ✅ Auto-generating policy files on scaffold is welcome
- ❌ Maintaining controller templates is out of scope for Pundit

My plan is to take a surgical approach — only hook into the scaffold generator to auto-generate policy files (and their corresponding test files), without touching controller templates at all.

Would this direction work for you?

ponio-ponix:
Hi! I'm interested in working on this issue.
I noticed that #876 was closed without review. Is this still something the maintainers would like to see contributed?
If so, I'd be happy to work on the scoped approach discussed above.

PR #794, opened 2024-02-28
Add scaffold generators

hoverjet:
This PR introduces a Pundit generator for Rails scaffolds and adds `authorize` and `policy_scope` Pundit's methods to scaffolded controllers. It modifies the scaffold generator to include a hook for Pundit policies and overrides scaffold controller templates through a Railtie to add Pundit's methods.
Example scaffold command:
```
# rails g scaffold Post
      invoke  active_record
      create    db/migrate/20240225223237_create_posts.rb
      create    app/models/post.rb
      invoke    rspec
      create      spec/models/post_spec.rb
      invoke  resource_route
       route    resources :posts
      invoke  scaffold_controller
      create    app/controllers/posts_controller.rb
      invoke    erb
      create      app/views/posts
      create      app/views/posts/index.html.erb
      create      app/views/posts/edit.html.erb
      create      app/views/posts/show.html.erb
      create      app/views/posts/new.html.erb
      create      app/views/posts/_form.html.erb
      create      app/views/posts/_post.html.erb
      invoke    resource_route
      invoke    rspec
      create      spec/requests/posts_spec.rb
      create      spec/views/posts/edit.html.erb_spec.rb
      create      spec/views/posts/index.html.erb_spec.rb
      create      spec/views/posts/new.html.erb_spec.rb
      create      spec/views/posts/show.html.erb_spec.rb
      create      spec/routing/posts_routing_spec.rb
      invoke    helper
      create      app/helpers/posts_helper.rb
      invoke      rspec
      create        spec/helpers/posts_helper_spec.rb
      invoke    jbuilder
      create      app/views/posts/index.json.jbuilder
      create      app/views/posts/show.json.jbuilder
      create      app/views/posts/_post.json.jbuilder
      invoke  policy
      create    app/policies/post_policy.rb
      invoke    rspec
      create      spec/policies/post_policy_spec.rb
```
Generated `PostsController`:
```
class PostsController < ApplicationController
  before_action :set_post, only: %i[show edit update destroy]

  # GET /posts
  def index
    @posts = policy_scope(Post.all)
  end

  # GET /posts/1
  def show
    authorize @post
  end

  # GET /posts/new
  def new
    @post = Post.new
    authorize @post
  end

  # GET /posts/1/edit
  def edit
    authorize @post
  end

  # POST /posts
  def create
    @post = Post.new(post_params)
    authorize @post

    if @post.save
      redirect_to @post, notice: "Post was successfully created."
    else
      render :new, status: :unprocessable_entity
    end
  end

  # PATCH/PUT /posts/1
  def update
    authorize @post
    if @post.update(post_params)
      redirect_to @post, notice: "Post was successfully updated.", status: :see_other
    else
      render :edit, status: :unprocessable_entity
    end
  end

  # DELETE /posts/1
  def destroy
    authorize @post
    @post.destroy!
    redirect_to posts_url, notice: "Post was successfully destroyed.", status: :see_other
  end

  private
    # Use callbacks to share common setup or constraints between actions.
    def set_post
      authorize @post
    end

    # Only allow a list of trusted parameters through.
    def post_params
      params.fetch(:post, {})
    end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,21 @@ Use the supplied generator to generate policies:
 ``` sh
 rails g pundit:policy post
 ```
+The scaffold command:
+```sh
+rails g scaffold Post
+```
+will also generate a policy and a scaffolded controller with Pundit's authorization method calls for actions.
+```sh
+...
+      invoke  scaffold_controller
+      create    app/controllers/posts_controller.rb
+...
+      invoke  policy
+      create    app/policies/post_policy.rb
+      invoke    rspec
+      create      spec/policies/post_policy_spec.rb
+```
 
 ## Closed systems
 
--- a/lib/generators/pundit/scaffold/scaffold_generator.rb
+++ b/lib/generators/pundit/scaffold/scaffold_generator.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+require "rails/generators/rails/scaffold/scaffold_generator"
+
+module Rails
+  module Generators
+    class ScaffoldGenerator
+      hook_for :policy, in: :pundit, type: :boolean, default: true
+    end
+  end
+end
--- a/lib/generators/rails/scaffold_controller/templates/api_controller.rb.tt
+++ b/lib/generators/rails/scaffold_controller/templates/api_controller.rb.tt
@@ -1,0 +1,61 @@
+<% module_namespacing do -%>
+class <%= controller_class_name %>Controller < ApplicationController
+  before_action :set_<%= singular_table_name %>, only: %i[ show update destroy ]
+
+  # GET <%= route_url %>
+  def index
+    @<%= plural_table_name %> = policy_scope(<%= orm_class.all(class_name) %>)
+    render json: @<%= plural_table_name %>
+  end
+
+  # GET <%= route_url %>/1
+  def show
+    authorize @<%= singular_table_name %>
+    render json: @<%= singular_table_name %>
+  end
+
+  # POST <%= route_url %>
+  def create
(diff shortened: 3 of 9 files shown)


Burgestrand:
I believe this introduces a dependency on Rails? Pundit should not have Rails as a dependency.

hoverjet:
I hope that by adding the condition `require "pundit/railtie" if defined?(Rails)`, it is ensured that the gem can be used in both Rails and non-Rails Ruby environments. This makes the inclusion of the Railtie and generators optional and dependent on the runtime context.

Burgestrand:
> I hope that by adding the condition `require "pundit/railtie" if defined?(Rails)`, it is ensured that the gem can be used in both Rails and non-Rails Ruby environments. This makes the inclusion of the Railtie and generators optional and dependent on the runtime context.

I believe you're correct, that seems like it's the recommended approach when viewing the documentation: https://api.rubyonrails.org/classes/Rails/Railtie.html

I'm for the spirit of more guidance in how to use Pundit, and making it more convenient.

However, I'm significantly more hesitant towards providing and maintaining controller generation templates. That's a lot of code to have an opinion on, when Pundit is mostly only concerned with a relatively small part of the generated code. 

My gut feeling is that I'd like a more surgical approach. I'm not sure that's possible.

Regardless I feel I also need to run this PR by another maintainer friend before making a decision.

Burgestrand:
I also missed that this PR was born out of https://github.com/varvet/pundit/issues/735

It's worth mentioning that I still think that generating _policies_ automatically on scaffold is a good idea. Pundit kind of own policies in that sense, and can have opinions on how to write those.

Burgestrand:
Thank you for your work. I've decided that maintaining controller templates is not something for Pundit to do.

PR #797, opened 2024-03-01
First pass of Pundit::Context

Burgestrand:
See https://github.com/varvet/pundit/discussions/774 initial draft/rationale.

## Caution
* We're introducing a new method [`Pundit::Authorization#pundit`](https://github.com/varvet/pundit/blob/d7251544dc552b7fe9f95933882279ae37e566ab/lib/pundit/authorization.rb#L19-L24). This is a namespace that is technically not ours, as this module included in controllers.

## Hopefully this allows us to:
* ... improve/fix existing functionality while staying backwards-compatible, by using [dependency injection](https://github.com/varvet/pundit/blob/d7251544dc552b7fe9f95933882279ae37e566ab/lib/pundit/authorization.rb#L22).
* ... introduce new functionality without necessarily bloating controllers in `Pundit::Authorization`.
* ... introduce an actual API for namespaces rather than overriding `authorize`/`policy`.

### DIFF
--- a/lib/pundit/context.rb
+++ b/lib/pundit/context.rb
@@ -1,0 +1,118 @@
+# frozen_string_literal: true
+
+module Pundit
+  class Context
+    def initialize(user:, policy_cache: {})
+      @user = user
+      @policy_cache = policy_cache
+    end
+
+    attr_reader :user
+
+    # @api private
+    attr_reader :policy_cache
+
+    # Retrieves the policy for the given record, initializing it with the
+    # record and user and finally throwing an error if the user is not
+    # authorized to perform the given action.
+    #
+    # @param user [Object] the user that initiated the action
+    # @param possibly_namespaced_record [Object, Array] the object we're checking permissions of
+    # @param query [Symbol, String] the predicate method to check on the policy (e.g. `:show?`)
+    # @param policy_class [Class] the policy class we want to force use of
+    # @raise [NotAuthorizedError] if the given query method returned false
+    # @return [Object] Always returns the passed object record
+    def authorize(possibly_namespaced_record, query:, policy_class:)
+      record = pundit_model(possibly_namespaced_record)
+      policy = if policy_class
+        policy_class.new(user, record)
+      else
+        policy_cache[possibly_namespaced_record] ||= policy!(possibly_namespaced_record)
+      end
+
+      raise NotAuthorizedError, query: query, record: record, policy: policy unless policy.public_send(query)
+
+      record
+    end
+
+    # Retrieves the policy scope for the given record.
+    #
+    # @see https://github.com/varvet/pundit#scopes
+    # @param user [Object] the user that initiated the action
+    # @param scope [Object] the object we're retrieving the policy scope for
+    # @raise [InvalidConstructorError] if the policy constructor called incorrectly
+    # @return [Scope{#resolve}, nil] instance of scope class which can resolve to a scope
+    def policy_scope(scope)
+      policy_scope_class = policy_finder(scope).scope
+      return unless policy_scope_class
+
+      begin
+        policy_scope = policy_scope_class.new(user, pundit_model(scope))
+      rescue ArgumentError
+        raise InvalidConstructorError, "Invalid #<#{policy_scope_class}> constructor is called"
+      end
+
+      policy_scope.resolve
+    end
+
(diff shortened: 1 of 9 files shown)


Burgestrand:
Right now this change is backwards-compatible, possibly with the exception of our new `pundit` method. We could probably try this out in the wild, and then expand on the `Pundit::Context` API surface to introduce namespacing.

lib/pundit/context.rb
```diff
@@ -0,0 +1,127 @@
...
+    def initialize(user:, policy_cache: {}, scope_cache: {})
+      @user = user
+
+      @policy_cache = policy_cache
+      @scope_cache = scope_cache
+    end
+
+    attr_reader :user
+
+    def with_user(new_user)
```
Burgestrand, line 14: Not using this yet. Let's remove it.

lib/pundit/context.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+module Pundit
+  class Context
+    def initialize(user:, policy_cache: {}, scope_cache: {})
+      @user = user
+
+      @policy_cache = policy_cache
```
Burgestrand, line 8: These are technically only here for backwards-compatibility.

Burgestrand:
Unless it's far too controversial I'd like to be able to remove the singleton methods on `Pundit` class in the future, and instead call them through a `Pundit::Context`. This is only if you're using Pundit outside of Rails.

For those who uses `Pundit::Authorization` (i.e. Rails), I'm thinking not much would change.

lib/pundit/context.rb
```diff
@@ -0,0 +1,127 @@
...
+    # @param user [Object] the user that initiated the action
+    # @param possibly_namespaced_record [Object, Array] the object we're checking permissions of
+    # @param query [Symbol, String] the predicate method to check on the policy (e.g. `:show?`)
+    # @param policy_class [Class] the policy class we want to force use of
+    # @raise [NotAuthorizedError] if the given query method returned false
+    # @return [Object] Always returns the passed object record
+    def authorize(possibly_namespaced_record, query:, policy_class:)
+      record = pundit_model(possibly_namespaced_record)
+      policy = if policy_class
+        policy_class.new(user, record)
```
mattzollinhofer, line 37: In my case I think I'd also like to add the `policy_class` case to the cache and prefer the cache when using `policy`. PR coming your way to show what I'm talking about.

The question may come down to if you agree and how pundit wants to view what it means to specify a `policy_class`.

lib/pundit/context.rb
```diff
@@ -0,0 +1,127 @@
...
+      begin
+        policy_scope = policy_scope_class.new(user, pundit_model(scope))
+      rescue ArgumentError
+        raise InvalidConstructorError, "Invalid #<#{policy_scope_class}> constructor is called"
+      end
+
+      policy_scope.resolve
+    end
+
+    # Retrieves the policy scope for the given record.
```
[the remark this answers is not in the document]
Burgestrand: I really like changes like these. They improve things for everyone.

For me it can be really hard to spot these improvements since I'm somewhat comfortable in the codebase already, so I appreciate it.

Burgestrand:
@mattzollinhofer I've added some more commits to kinda hint at where I'm going at.

A change I haven't gotten to is passing _more_ information to `policy_cache.fetch`, at least the `user` and maybe the passed-in `policy_class`. Once that's done it would be up to the cache strategy to turn those three into a cache key.

We would achieve backwards-compatibility by having a `LegacyCache` as default cache, and that cache simply ignores the `user` and `policy_class` passed. It would then work just as before.

By then implementing a new cache strategy that's opt-in we can get caching behaviour when passing `policy_class`. How that looks then later in the view I'm not sure. Maybe we would require passing the overridden policy from the view too, so that the cache key remains the same - otherwise it would be impossible to mix explicit `policy_class` with pundit-found policy.

Burgestrand:
Another idea I had for this is some shortcuts to how one could override the default behavior in specific controllers, e.g:
```ruby
class AdminControllerBase < ApplicationController
  def pundit = super.with_namespace(Admin)
end
```

Burgestrand:
Had a few people at RailsConf try this out in their apps, and nothing broke! Thank you all :)

PR #801, opened 2024-03-21
Considering integrating the Pundit::Context even more

mattzollinhofer:
Prefer `policy_cache` if available to `policy_finder`. 

I'm unsure if there are other implications for this, but for the use
case of sharing context between the typical controller use (`authorize`)
vs the view typical use (`policy`) this adds value. Without this the
`authorize` and `policy` methods are still disconnected as far as I can
tell.

It looks like you were thoughtful about only choosing the else clause to
be stored in the cache, so I might be missing a consideration you were
making. But, I'd argue that if the client specifies a policy_class for
a given record then the goal is for that policy_class to apply to all
records through the pundit-based interactions. It was our intention at
least.

### DIFF
--- a/lib/pundit/context.rb
+++ b/lib/pundit/context.rb
@@ -34,7 +34,7 @@ module Pundit
     def authorize(possibly_namespaced_record, query:, policy_class:)
       record = pundit_model(possibly_namespaced_record)
       policy = if policy_class
-        policy_class.new(user, record)
+        policy_cache[possibly_namespaced_record] ||= policy_class.new(user, record)
       else
         policy_cache[possibly_namespaced_record] ||= policy!(possibly_namespaced_record)
       end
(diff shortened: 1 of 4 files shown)


lib/pundit/context.rb
```diff
@@ -108,7 +108,7 @@ def policy(record)
     # @raise [InvalidConstructorError] if the policy constructor called incorrectly
     # @return [Object] instance of policy class with query methods
     def policy!(record)
-      policy = policy_finder(record).policy!
+      policy = policy_cache[record] || policy_finder(record).policy!
```
mattzollinhofer, line 111: If you agree with this mentality, I wonder if this could be taken a step further by moving this or similar logic into the `def policy_finder` private method.

Burgestrand:
Thanks!

Yes, I've been thinking about this. I believe that the policy and scope cache _should_ apply when you supply a `policy_class`.

However, changing the current behaviour is not backwards-compatible: adding caching behaviour to a method where there used to be none.

I believe there's a way forward where caching like this can be opt-in behaviour, while keeping backwards-compatibility by default. Probably by adjusting the dependencies injected into `Pundit::Context` somehow.

PR #806, opened 2024-05-08
Add customizable permit matcher description

### LINKED ISSUE #759: Enable custom description for permit matcher

gregfletch:
Updates #760 by addressing feedback provided on the original PR.

Closes #759

### DIFF
--- a/README.md
+++ b/README.md
@@ -783,6 +783,14 @@ describe PostPolicy do
 end
 ```
 
+You can customize the description used for the `permit` matcher:
+
+``` ruby
+Pundit::RSpec::Matchers.description do
+  "permit the user"
+end
+```
+
 An alternative approach to Pundit policy specs is scoping them to a user context as outlined in this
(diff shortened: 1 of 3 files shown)


Burgestrand:
Hi again Greg! I'm finally back home now ☺️ 

This looks good! I realise we lack documentation of the proc-variation of the description, but that can be adjusted later so I'll merge this right away 🙂

PR #812, opened 2024-06-10
Add the policy class to the `NotAuthorizedError` message

smitssjors:
This PR fixes #681 and #647.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- 
+
 ## 2.3.2 (2024-05-08)
 
 - Refactor: First pass of Pundit::Context (#797)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-- 
+- Improve the `NotAuthorizedError` message to include the policy class.
+Furthermore, in the case where the record passed is a class instead of an instance, the class name is given.
 
 ## 2.3.2 (2024-05-08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-- Improve the `NotAuthorizedError` message to include the policy class. (#812)
-Furthermore, in the case where the record passed is a class instead of an instance, the class name is given.
+- Improve the `NotAuthorizedError` message to include the policy class.
+Furthermore, in the case where the record passed is a class instead of an instance, the class name is given. (#812)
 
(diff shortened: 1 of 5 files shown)


Burgestrand:
Hi! Thanks for updating the old PR!

This does _not_ fix #647, but it does fix #681.

I'm not sure why the pull request templates we have didn't get proposed, but please have a look at it: https://github.com/varvet/pundit/blob/main/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md - this PR is missing a changelog entry 🙂

smitssjors:
I updated the PR to now actually take care of #647 and I updated the `CHANGELOG.md`

I did not get the template when I made the PR via the button in my fork. I did however read `CONTRIBUTING.md` so I added updating the CHANGELOG as a step to hopefully prevent these mistakes in the future.

Burgestrand:
The jruby failures are unrelated to this change. Something is up with the more recent jruby version on our GitHub CI pipeline.

Burgestrand:
Either way I think this should probably be good! I'll most likely merge this soon. Travelling to Baltic Ruby today though 😊

lib/pundit.rb
```diff
@@ -36,7 +36,9 @@ def initialize(options = {})
         @record = options[:record]
         @policy = options[:policy]
 
-        message = options.fetch(:message) { "not allowed to #{query} this #{record.class}" }
+        record_name = record.is_a?(Class) ? record.name : "this #{record.class}"
+
+        message = options.fetch(:message) { "not allowed to #{policy.class}##{query} #{record_name}" }
```
Burgestrand, line 41: Hi! I've thought about this some more and I've got a change that's worth making.

The reasons for the change are:
1. `record.to_s` works better for anonymous classes, and it's what we use in PolicyFinder so it's somewhat battle-tested.
2. If there's an explicit message we should not calculate a record name.

```suggestion
        message = options.fetch(:message) do
          record_name = record.is_a?(Class) ? record.to_s : "this #{record.class}"
          "not allowed to #{policy.class}##{query} #{record_name}"
        end
```

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Improve the `NotAuthorizedError` message to include the policy class.
```
Burgestrand, line 5: Please add the issue number here :)

```suggestion
- Improve the `NotAuthorizedError` message to include the policy class. (#812)
```

Burgestrand (changes requested):

Burgestrand (approved):

Burgestrand:
Thank you!

Issue #816, opened 2024-06-25
Support :focus for rspec permissions blocks


pas256:
When writing tests using rspec, there is a nice helper called `permissions` that works similar to `describe`, but it does not support the `:focus` attribute used to limit which tests are run.

Example:

```ruby
RSpec.describe SheetPolicy, type: :policy do
  permissions :index? do
    it 'grants access if the workbook is created by the user' do
      expect(SheetPolicy).to permit(user, create(:sheet, workbook:))
    end
  end

  permissions :show?, :focus do
    it 'grants access if the workbook is created by the user' do
      expect(SheetPolicy).to permit(user, create(:sheet, workbook:))
    end
  end
```

When you have `config.filter_run_when_matching :focus` in your `spec_helper.rb` file, it should only run the `index` block of tests, but from what I can tell, `permissions` does not support it.

Burgestrand:
Without looking into the details too much here, this feels something we should reasonably support.

Burgestrand:
This was fixed in #820!

PR #820, opened 2024-07-13
Support :focus for rspec permissions blocks

StevenElberger:
I thought this issue would be nice to have so I took a stab at it:
https://github.com/varvet/pundit/issues/816

Please let me know if there's anything missing!

### DIFF
--- a/spec/dsl_spec.rb
+++ b/spec/dsl_spec.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe "Pundit RSpec DSL" do
+  let(:fake_rspec) do
+    double = class_double(RSpec::ExampleGroups)
+    double.extend(::Pundit::RSpec::DSL)
+    double
+  end
+  let(:block) { proc { "block content" } }
+
+  it "calls describe with the correct metadata and without :focus" do
+    expected_metadata = { permissions: %i[item1 item2], caller: instance_of(Array) }
+    expect(fake_rspec).to receive(:describe).with("item1 and item2", match(expected_metadata)) do |&block|
+      expect(block.call).to eq("block content")
+    end
+
+    fake_rspec.permissions(:item1, :item2, &block)
+  end
+
+  it "calls describe with the correct metadata and with :focus" do
+    expected_metadata = { permissions: %i[item1 item2], caller: instance_of(Array), focus: true }
+    expect(fake_rspec).to receive(:describe).with("item1 and item2", match(expected_metadata)) do |&block|
+      expect(block.call).to eq("block content")
+    end
+
+    fake_rspec.permissions(:item1, :item2, :focus, &block)
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Improve the `NotAuthorizedError` message to include the policy class.
 Furthermore, in the case where the record passed is a class instead of an instance, the class name is given. (#812)
 - Add customizable permit matcher description (#806)
+- Add support for filter_run_when_matching :focus with permissions helper. (#820)
 
 ## 2.3.2 (2024-05-08)
 
--- a/README.md
+++ b/README.md
@@ -821,6 +821,14 @@ PostPolicy
     is expected to permit user with role admin to access record with ID 130
 ```
 
+### Focus Support
+
+If your RSpec config has `filter_run_when_matching :focus`, you may tag the `permissions` helper like so:
+
+```
+permissions :show?, :focus do
+```
+
 ### Scope Specs
 
 Pundit does not provide a DSL for testing scopes. Test them like you would a regular Ruby class!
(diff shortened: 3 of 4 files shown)


Burgestrand:
Spontaneously this looks good! I'm currently enjoying Swedish vacation so I'll take a proper look in august 😊

spec/dsl_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "spec_helper"
+
+RSpec.describe "Pundit RSpec DSL" do
+  before do
+    RSpec.configure do |config|
+      config.include ::Pundit::RSpec::DSL
+    end
```
Burgestrand, line 7: I'm not keen on modifying global state in a `before`-filter. It'll leak to other tests, and we want every test to be isolated from others.

Might I suggest creating a double that mimics rspec behaviour here instead?

```ruby
let(:fake_rspec) do
  double = class_double(RSpec::ExampleGroups)
  double.extend(::Pundit::RSpec::DSL)
  allow(double).to receive(:describe).and_yield
  double
end
```

You would then be able to expect behaviour on the double (note using [yield_control](https://rspec.info/features/3-13/rspec-expectations/built-in-matchers/yield/) matcher here, too):

```ruby
expect { |block| fake_rspec.permissions(..., &block) }.to yield_control
expect(fake_rspec).to have_received(:describe).with(...)
```

spec/dsl_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  let(:block) { proc { "block content" } }
+
+  it "calls describe with the correct metadata and without :focus" do
+    expected_metadata = { permissions: %i[:item1, :item2], caller: instance_of(Array) }
+    expect(self).to receive(:describe).with("item1 and item2", hash_including(expected_metadata)) do |&block|
```
Burgestrand, line 14: Careful here, `hash_including` is lenient when it comes to _additional_ content. If the implementation always added `focus: true` then this test would still pass, and so would the other one.

Instead of `hash_including` we want `match` here instead.

spec/dsl_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  before do
+    RSpec.configure do |config|
+      config.include ::Pundit::RSpec::DSL
+    end
+  end
+
+  let(:block) { proc { "block content" } }
+
+  it "calls describe with the correct metadata and without :focus" do
+    expected_metadata = { permissions: %i[:item1, :item2], caller: instance_of(Array) }
```
Burgestrand, line 13: Test failures probably already caught this, but `%i` shouldn't have a comma `,` nor colon `:` 🙂

Burgestrand (changes requested):
Thank you! I think the implementation is good! Readme needs some tiny adjustments, and the test needs a few changes too. Please see my comments 😊

StevenElberger:
@Burgestrand thank you so much for your very helpful and insightful feedback! 🙇  I hope the latest changes satisfy your comments but if not please let me know!

spec/dsl_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  let(:fake_rspec) do
+    double = class_double(RSpec::ExampleGroups)
+    double.extend(::Pundit::RSpec::DSL)
+    allow(double).to receive(:describe).and_yield
+    double
+  end
+  let(:block) { proc { "block content" } }
+
+  it "calls describe with the correct metadata and without :focus" do
+    expect { |block| fake_rspec.permissions(:item1, :item2, &block) }.to yield_control
```
Burgestrand, line 13: Thanks for making the changes, it's looking good!

I have but one small nitpick left, which I myself caused by not being clear enough in my previous comment. Sorry!

I believe you should remove line 7 (`allow(double).to receive(:describe)`), line 13 (`expect { ... }.to yield_control`), and line 24 (`expect { ... }.to yield_control` }). They don't add anything to the test, and they make for more reading material.

It's not strictly important why I suggested these, but I'll explain my reasoning for the curious:
* Line 7 allows us to change `expect(fake_rspec).to receive(:describe)` into `expect(fake_rspec).to have_received(:describe)`, which means we can collect all assertions at the end of the test. I find the test writing style of "1. setup, 2. execute, 3. assert" to work really well in general, so I gravitate towards that.
* Line 13 and 24 were intended as an alternative to the bare `fake_rspec.permissions(...)`, but honestly I think asserting the block is called in our expectation around `describe` is more useful.

Burgestrand (changes requested):

spec/dsl_spec.rb
StevenElberger: @Burgestrand sorry for misunderstanding your suggestions and thank you for clarifying! I really enjoyed the explanations. hopefully these latest changes are satisfactory! 😅

Burgestrand (approved):
Thank you, this looks great!

Burgestrand:
@StevenElberger rubocop caught a missing comment! Once you add that I believe this is ready to merge 😊

StevenElberger:
> @StevenElberger rubocop caught a missing frozen string literal comment! Once you add that I believe this is ready to merge 😊

@Burgestrand got it! should be good to go now, thank you! 😃

Burgestrand:
@StevenElberger merged, thank you! I'll see if I can get a release going with this change soon 😊

PR #825, opened 2024-08-18
Optionally allow overriding the policy class for permitted params

Numie:
Fixes https://github.com/varvet/pundit/issues/742


Burgestrand:
Hi, thanks for the pull request!

First things first, I adjusted the PR description to use the PR template we've got going on. This change would need a CHANGELOG entry.

I feel the scope of this PR is somewhat inflated by the moving of the `permitted_attributes` implementation into `Pundit::Context`, but I'm not sure why that was done. I think that can be skipped entirely, to keep the change small.

I also notice that `pundit_params_for(record)` also looks up a policy, and I wonder if that policy shouldn't be _the same policy_ if you pass in `policy_class`. That would necessitate a change there too.

Numie:
> Hi, thanks for the pull request!
> 
> I adjusted the PR description to use the PR template we've got going on. This change would need a CHANGELOG entry.
> 
> On another note, I feel the scope of this PR is somewhat inflated by the moving of the `permitted_attributes` implementation into `Pundit::Context`, but I'm not sure why that was done. I think that can be skipped entirely, to keep the change small.
> 
> I also notice that `pundit_params_for(record)` also looks up a policy, and I wonder if that policy shouldn't be _the same policy_ if you pass in `policy_class`. That would necessitate a change there too.
> 
> I'll ponder the whole `Pundit::Context` and permitted parameters relationship in the meanwhile, too.

Hi @Burgestrand,

No problem re: a CHANGELOG entry.

Re: moving of the `permitted_attributes` implementation into `Pundit::Context`, I was following the pattern implemented for `Pundit::Authorization#authorize` and `Pundit::Authorization#policy_scope`, where the `policy_class` arg is forwarded to a method of the same name on `Pundit::Context`. My understanding is that this is necessary to initialize a new policy class because only `Pundit::Context` has access to the user.

Let me know your thoughts!

Issue #811, opened 2024-06-10
Pundit caching isn't sensitive to current user changing


danielfone:
@Burgestrand Hi there! Firstly, thanks for all the hard work maintaining pundit, it's a brilliant gem and I'm a long time fan.

Re https://github.com/varvet/pundit/pull/797

> This should be backwards-compatible, so let me know if this breaks anything.

Unfortunately, 2.3.2 has broken a key piece of authorization functionality in one of my apps. I'm not sure if it's because I was doing it wrong in the first place, or if this is an unintentional regression. I have a controller concern that allows admins to impersonate/become another user, to view the app from another user's perspective. The way we implement this is approximately:

```ruby
module AdminOverride
  extend ActiveSupport::Concern

  included do
    before_action :override_with_admin_user, if: -> { session[:admin_user_as] }
  end

  def override_with_admin_user
    return unless current_user

    authorize :user, :become? # First, ensure that the current user has permissions to 'become' another user
    @original_user = current_user
    @current_user = User.find session[:admin_user_as] # If so, override the current user
  end

end
```

Unfortunately, since `Pundit::Context` is initialised on the first policy look up (when the current user is the admin) and memoizes the current user at that point, all subsequent policies are checked against the admin user, and not the impersonated user.

I can work around it in a variety of ways, but wanted to let you know and see if this is something you're concerned about. It was unfortunate that it (a) happened in a patch release, and (b) my integration testing wasn't sufficient to catch this change, since it slipped into production and I had a _very_ panicked phone call from an admin who thought users could now see everything. 😂  You live and learn!

Thanks again.

Burgestrand:
Hi @danielfone this is amazing feedback!

I'll ponder what Pundit should do here. The context cache might benefit from being current user-sensitive, to avoid accidents. It also might be worth having a recommended approach documented in the README.

Not sure how much Pundit itself should help with the multi-user case, mostly because it's a relatively rare use-case.

Also, **one very important thing to consider**, is that previous releases of Pundit _did_ also have issues with caching and changing the current user. We've been caching policy lookups for years, and that cache is also not sensitive to the current user changing.

Effectively this won't be good:
```ruby
def show?
  post = Post.find(...)
  # current_user = post.author here
  authorize post

  @current_user = User::Guest.new
  authorize post # => WARNING! cached post policy from earlier
end
``` 

So if you do switch users like this, you need to be mindful of all the caches: `pundit`, `policies`, and `policy_scopes`.

[...]

PR #830, opened 2024-09-13
add hook for clear pundit context

furkanural:
Fix https://github.com/varvet/pundit/issues/811

As far as I can see, the caches are not being cleared after a `UserContext` change. Therefore, policies are being performed using the old UserContext. I would like to implement a simple solution to this issue. With this solution, caches can be cleared when deemed necessary.

```ruby
class ApplicationController
  include Pundit::Authorization
  before_action :switch_user, if: :should_switch_user?

  def switch_user
    current_user = User.find(params[:user_id])
    pundit_reset!  # Ensure that the Pundit context is reset for the new user
  end
end
```

## To do

### DIFF
--- a/README.md
+++ b/README.md
@@ -669,7 +669,7 @@ class ApplicationController
 
   def switch_account
     set_current_account(Account.find(params[:account_id]))
-    clear_pundit_context!
+    pundit_reset!
   end
 
   def pundit_user
--- a/README.md
+++ b/README.md
@@ -660,6 +660,24 @@ class ApplicationController
 end
 ```
 
+If you need to change the context for any reason, you will need to clear the caches stored in the context. You can use the hook below to do this.
+
+```ruby
+class ApplicationController
+  include Pundit::Authorization
+  before_action :switch_account, if: :should_switch_account?
+
+  def switch_account
+    set_current_account(Account.find(params[:account_id]))
+    clear_pundit_context!
+  end
+
+  def pundit_user
+    UserContext.new(current_user, current_account)
+  end
+end
+```
+
 ## Strong parameters
 
 In Rails,
--- a/README.md
+++ b/README.md
@@ -591,11 +591,11 @@ To handle user switching, you can use the following pattern in your controller:
 ```ruby
 class ApplicationController
   include Pundit::Authorization
-  before_action :switch_user, if: :should_switch_user?
 
-  def switch_user
-    current_user = User.find(params[:user_id])
-    pundit_reset!  # Ensure that the Pundit context is reset for the new user
+  def switch_user_to(user)
+    terminate_session if authenticated?
+    start_new_session_for user
+    pundit_reset!
   end
 end
 ```
--- a/README.md
+++ b/README.md
@@ -660,20 +660,20 @@ class ApplicationController
 end
 ```
 
-If you need to change the context for any reason, you will need to clear the caches stored in the context. You can use the hook below to do this.
+If you need to change the context for any reason, you will need to clear the caches stored in the Pundit. You can use the hook below to do this.
 
 ```ruby
 class ApplicationController
   include Pundit::Authorization
-  before_action :switch_account, if: :should_switch_account?
+  before_action :switch_user, if: :should_switch_user?
 
   def switch_account
-    set_current_account(Account.find(params[:account_id]))
+    set_current_user(User.find(params[:user_id]))
     pundit_reset!
   end
 
   def pundit_user
-    UserContext.new(current_user, current_account)
+    UserContext.new(current_user, request.ip)
   end
 end
 ```
--- a/README.md
+++ b/README.md
@@ -582,6 +582,25 @@ def pundit_user
   User.find_by_other_means
 end
 ```
+### Handling User Switching in Pundit
+
+When switching users in your application, it's important to reset the Pundit user context to ensure that authorization policies are applied correctly for the new user. Pundit caches the user context, so failing to reset it could result in incorrect permissions being applied.
+
+To handle user switching, you can use the following pattern in your controller:
+
+```ruby
+class ApplicationController
+  include Pundit::Authorization
+  before_action :switch_user, if: :should_switch_user?
+
+  def switch_user
+    current_user = User.find(params[:user_id])
+    pundit_reset!  # Ensure that the Pundit context is reset for the new user
+  end
+end
+```
+
+Make sure to invoke `pundit_reset!` whenever changing the user. This ensures the cached authorization context is reset, preventing any incorrect permissions from being applied.
 
 ## Policy Namespacing
 In some cases it might be helpful to have multiple policies that serve different contexts for a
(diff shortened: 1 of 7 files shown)


Burgestrand:
Thank you for the PR! I like the implementation. Clear, well-written, and tests. Nice!

I have some hesitations that I need to ponder:
* Against: I don't believe this is a common need for most applications.
* Against: It expands public API of Pundit, which needs to be carefully considered and maintained.
* For: Having an official reset allows people using Pundit to trust this to be future-proof. That has value.

Before this would be merged, assuming it will be merged, I think some changes should be made:
* Documentation change mentions "accounts", but we're dealing with `user` in Pundit. The hypothetical method should probably be `switch_user` to keep things simple.
* I believe the method name should be `pundit_reset`, to follow the convention.
* I'm not certain of this, but we should _probably_ clear the flags too.
  * `@_pundit_policy_authorized`
  * `@_pundit_policy_scoped`

Burgestrand (changes requested):
https://github.com/varvet/pundit/pull/830#issuecomment-2399124211

furkanural:
@Burgestrand Thank you for your thoughtful feedback! I appreciate your comments and am glad the implementation resonates well with you.

Regarding the concerns:
**Against: Common need for most applications:** I understand that this might not be a widely requested feature. However, I believe there are certain cases where having a reliable reset mechanism would greatly benefit applications with complex authorization flows. That said, I see your point and think it’s important to weigh the demand versus maintenance cost.

**Against: Expands public API:** You’re absolutely right :)

**For: Future-proofing:** Agreed! One of the key advantages of adding this feature is precisely the long-term value it provides for those who rely on Pundit. It gives them confidence that their systems will remain stable and maintainable as the library evolves.

Now, for the specific changes:
- **Documentation and naming:** Excellent catch! I agree that the documentation should refer to "users" rather than "accounts" to maintain consistency within Pundit’s terminology. I'll update the method name to `switch_user` as well—it keeps things intuitive.
- **Method naming convention:** I’ve renamed the method to `pundit_reset!` to follow Ruby conventions for methods that perform a significant change or could have side effects. Since the reset operation clears important state, I think the bang method makes sense here.
- **Flag clearing:** I see your point about potentially clearing the flags. I’ll investigate this further to ensure we aren’t leaving any unnecessary state behind after a reset. I'll review both `@_pundit_policy_authorized` and `@_pundit_policy_scoped` and make the necessary adjustments if needed.

Thanks again for the detailed review! I’ll proceed with the suggested changes and address the concerns you’ve raised.

lib/pundit/authorization.rb
```diff
@@ -167,6 +167,18 @@ def pundit_user
       current_user
     end
 
+    # Hook method which allows to clear the Pundit context.
+    #
+    # @see https://github.com/varvet/pundit#additional-context
```
Burgestrand, line 172: I believe this should explain the purpose of this method, the concept of "Pundit context" isn't well-established enough to refer to.

More specifically it might be helpful to know _why_ one would call this. It might be prudent to specifically call out "when you change the `current_user`".

Additionally, the link here has no relevance at all to the method. That should be removed.

spec/authorization_spec.rb
```diff
@@ -271,4 +271,38 @@ def to_params(*args, **kwargs, &block)
...
+
+    it "clears the current user" do
+      expect(controller.pundit.user).to eq controller.current_user
+
+      controller.current_user = new_user
+      expect(controller.pundit.user).not_to eq controller.current_user
+
+      controller.pundit_reset!
+      expect(controller.pundit.user).to eq controller.current_user
+    end
```
Burgestrand, line 286: This is something I missed my last review. Apologies!

I think the spirit of this test is good, it captures the intent. I believe it's worth adding one more test an abstraction level above it, to capture the very important property that _authorize_ itself also resets.

However, `#user` is kinda unofficially public API so we should be sure to test that too.

```suggestion
    it "allows authorize to react to a user change" do
      expect(controller.authorize(post)).to be_truthy

      controller.current_user = double
      controller.pundit_reset!
      expect { controller.authorize(post) }.to raise_error(Pundit::NotAuthorizedError)
    end
    
    it "clears the pundit context user" do
      expect(controller.pundit.user).to be(user)
      
      new_user = double
      controller.current_user = new_user
      expect { controller.pundit_reset! }.to change { controller.pundit.user }.from(user).to(new_user)
    end
```

Finally, a code style nitpick. The `let(:new_user)` is only used in a single test here, but the `let` makes me believe it isn't. It's not necessary, it should be inlined in the test.

spec/authorization_spec.rb
```diff
@@ -271,4 +271,38 @@ def to_params(*args, **kwargs, &block)
...
+
+    it "clears the policy cache" do
+      controller.authorize(post)
+      expect(controller.policies).not_to be_empty
+      expect(controller.pundit_policy_authorized?).to be true
```
Burgestrand, line 291: I understand how this assertion ended up in this test, but I believe it's out of place.

`pundit_policy_authorized?` (and `pundit_policy_scoped?`) doesn't have anything to do with the cache, it's a flag set by `authorize` to help catch errors in development: https://github.com/varvet/pundit?tab=readme-ov-file#ensuring-policies-and-scopes-are-used

As such, I believe these two methods should have their own tests using `#pundit_reset!`.

Burgestrand (changes requested):
Thank you for making the requested changes. I have some more 😁!

I've thought about this some more, and while I haven't yet finally decided I'm leaning even more into merging this.

README.md
```diff
@@ -660,20 +660,20 @@ class ApplicationController
...
-If you need to change the context for any reason, you will need to clear the caches stored in the context. You can use the hook below to do this.
+If you need to change the context for any reason, you will need to clear the caches stored in the Pundit. You can use the hook below to do this.
 
 ```ruby
 class ApplicationController
   include Pundit::Authorization
-  before_action :switch_account, if: :should_switch_account?
+  before_action :switch_user, if: :should_switch_user?
 
   def switch_account
```
Burgestrand, line 670: Missed this method name change 🙂

README.md
```diff
@@ -660,6 +660,24 @@ class ApplicationController
 end
 ```
 
+If you need to change the context for any reason, you will need to clear the caches stored in the Pundit. You can use the hook below to do this.
```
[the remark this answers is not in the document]
Burgestrand: Looking at this again, I think this method itself warrants it's own section. It shouldn't be under "Additional Context", maybe put it under "Customize pundit user" for now?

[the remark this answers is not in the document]
furkanural: Thank you for the detailed feedback! I’ve made the adjustments based on your suggestions, clarifying the method's purpose and improving the documentation. I appreciate your insights—they’ve really helped make things clearer. Let me know if there's anything else!

Burgestrand (commented):
I see my most review comments weren't properly sent. I'm sorry, I thought they were since github still says changes requested!

Regardless, here are the comments. Since these came in kinda late and I had some spare time, I took the liberty of rebasing your branch, and adjusting for my own recent comments. Will do another round of review, and maybe we can get this merged!

Burgestrand (approved):

Burgestrand:
@furkanural thank you, I merged this now!

PR #831, opened 2024-09-21
Prefer `require_relative` for internal requires

tagliala:
`require_relative` is preferred over `require` for files within the same project because it uses paths relative to the current file, making code more portable and less dependent on the load path.

This change updates internal requires to use `require_relative` for consistency, performance, and improved portability.

Refs:
- rubocop/rubocop#8748

## To do

- [x] ~I have added relevant tests.~
- [x] ~I have adjusted relevant documentation.~
- [x] ~I have added relevant lines to the CHANGELOG.~

---

I think this does not deserve an entry in the changelog

Relevant comment: https://github.com/rubocop/rubocop/issues/8748#issuecomment-2363327346

### DIFF
--- a/lib/pundit.rb
+++ b/lib/pundit.rb
@@ -1,16 +1,18 @@
 # frozen_string_literal: true
 
-require "pundit/version"
-require "pundit/policy_finder"
+require_relative "pundit/version"
+require_relative "pundit/policy_finder"
+
 require "active_support/concern"
 require "active_support/core_ext/string/inflections"
 require "active_support/core_ext/object/blank"
 require "active_support/core_ext/module/introspection"
 require "active_support/dependencies/autoload"
-require "pundit/authorization"
-require "pundit/context"
-require "pundit/cache_store/null_store"
-require "pundit/cache_store/legacy_store"
+
+require_relative "pundit/authorization"
+require_relative "pundit/context"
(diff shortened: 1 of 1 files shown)


Burgestrand:
Thank you for the PR. I read through the discussion in Rubocop. I didn't see any compelling arguments for why this would be better.

I'm of the opinion that relying on the load path is a Good Thing, specifically because it allows flexibility.

To be even more specific, if you for any reason wanted to prevent Pundit from loading e.g. `pundit/policy_finder` when we're using `require`, then you can do so. For example by creating `./pundit/policy_finder.rb` and then `$LOAD_PATH.unshift "."`. Why would you want to? I don't know, but it wouldn't be the first time for me.

Using `require_relative` would strip away this flexibility, but I don't see a benefit that compensates for that loss.

I'm open to hearing solid arguments about this, but without that this is a hard pass.

tagliala:
Thanks, can you share the concerns on the rubocop discussion? It may help others in deciding

Burgestrand:
> Thanks, can you share the concerns on the rubocop discussion? It may help others in deciding

That's a good suggestion, thank you!

Burgestrand:
https://github.com/rubocop/rubocop/issues/8748#issuecomment-2371501530

I wrote a comment. Thank you for the push!

PR #833, opened 2024-10-01
Register policies directories for Rails 8 code statistics

javierav:
This PR add a `Rails::Railtie` to `Pundit` for configure the new [stats directories](https://edgeapi.rubyonrails.org/classes/Rails/CodeStatistics.html) in upcomming Rails 8.


Burgestrand:
Thanks, this is neat! I need to find the time to test this out.

lib/pundit/railtie.rb
```diff
@@ -0,0 +1,21 @@
...
+        require "rails/code_statistics"
+
+        Rails::CodeStatistics.register_directory(
+          "Policies", Rails.root.join("app/policies").to_s
+        )
+
+        %w[test/policies spec/policies].each do |dir|
+          path = Rails.root.join(dir)
+
+          Rails::CodeStatistics.register_directory("Policies tests", path.to_s, test_directory: true) if path.exist?
```
Burgestrand, line 16: A _tiny_ nitpick, it seems the documentation you linked to has _singular_ for this, e.g. `Models` and `Model tests`.

javierav: done!

Burgestrand:
> @Burgestrand Can it be interesting to support older versions of rails? In such cases you would have to use a rake task (I never understood this ugly part of rails)

Gut feeling is "not really". I don't know how much you use `rails stats`, but for me it's probably less than once a year and it's not exactly an often-requested feature of Pundit 😄 

This PR change is small enough and uses documented APIs, which makes me comfortable merging it. Whatever we merge we'll have to support for a relatively long time.

Burgestrand:
Tested this on 8.0.1 and 7.2 and seems to work fine!

Burgestrand:
Thank you!

PR #850, opened 2025-02-28
Add documentation to provide a guide for how to use pundit with rails 8's built-in authentication generator

SuzukiRyuichiro:
Please disregard if this is not a necessary update on the documentation.
I was testing the combination of pundit and the new rails 8 authentication generator. If you use the built in authentication generator, you would get the currently authenticated user via `Current.user` not `current_user`. Since pundit expect `current_user`, some people suggesting to add that helper to the framework. rails/rails#54202

I just thought it might be a good thing to point out that it could be resolved with the portion that is already explained in the README.

My suggestion is just to make it more obvious by giving the exact solution.

## To do

- [ ] ~~I have added relevant tests.~~


Burgestrand:
Thank you, this looks great! I agree.

Issue #841, opened 2024-11-22
Support new params `expect` method to permit parameters


Linuus:
In Rails 8 the method `expect` has been introduced to mitigate some issues with the current `params.require(:foo).permit(:bar)` approach. One issue with the current approach is that if someone sends unexpected data, say `POST { foo: "bam" }` the Rails app will crash with a `NoMethodError` because the `.permit` method doesn't exist on `String`. The `expect` method handles this issue and will instead return a proper 400 error.

So, we should support this in upcoming versions of Pundit.

The `expect` method has a bit of a different syntax unfortunately so we can't just change it. I guess we either we have to make it required  > some version, or configurable in Pundit. I haven't looked closely on how to best handle this yet.

Docs: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect

Burgestrand:
This is probably my fastest reply ever.

Luckily we're using it in a very limited capacity, a single call-site:
https://github.com/varvet/pundit/blob/54cc64e45a3e4c606f555f05ce740e72c4f00946/lib/pundit/authorization.rb#L232-L238

Pundit should definitely support Rails 8.

Will ponder some more!

Burgestrand:
I misinterpreted your original message.

So if I understand correctly, behaviour doesn't change in an upgrade to Rails 8. However, `expect` might be the better option.

However, `expect` has an entirely different set of parameters, so a change would affect `permitted_attributes` too: https://github.com/varvet/pundit/blob/54cc64e45a3e4c606f555f05ce740e72c4f00946/lib/pundit/authorization.rb#L222-L230

It effectively would reduce to something more like:
```ruby
params.expect(PolicyFinder.new(record).param_key => policy.public_send(method_name))
```

... which I don't know if it works, but is indeed rather different and even makes `pundit_params_for` superfluous.

Burgestrand:
Two things important to me:
- Backwards-compatibility. We should not break it.

We can achieve backwards-compatibility by _adding_ to the API surface, e.g. `def expected_attributes(record, action:)` and then updating documentation.

However, what's important with that approach is that it's battle-tested before we recommend it everywhere. At the moment I'm not maintaining client projects with significant Pundit-usage (they don't need authorization), so I'm limited in my ability to battle-test it.

Burgestrand:
Kinda related to #742

Linuus:
[...]

PR #855, opened 2025-03-27
Add support for params.expect

Linuus:
Fixes: https://github.com/varvet/pundit/issues/841

## To do

- [ ] I have added relevant lines to the CHANGELOG.

This PR starts adding support for `params.expect` in Rails which is now recommended over `params.require.permit`.
See: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect

I'm a bit unsure what to do with `pundit_params_for` though. It's kind of redundant... but also kind of useful as mentioned in our Readme, if you want to fetch based on another "namespace" than the default. However, for expect it would essentially just turn into a wrapper method for `PolicyFinder.new(record).param_key`, so you can just override the `param_key` method on the Policy instead? 🤔

### DIFF
--- a/spec/authorization_spec.rb
+++ b/spec/authorization_spec.rb
@@ -273,6 +273,89 @@ describe Pundit::Authorization do
     end
   end
 
+  describe "#expected_attributes" do
+    it "checks policy for expected attributes" do
+      params = to_params(
+        post: {
+          title: "Hello",
+          votes: 5,
+          admin: true
+        }
+      )
+
+      action = "update"
+
+      expect(Controller.new(user, action, params).expected_attributes(post).to_h).to eq(
+        "title" => "Hello",
+        "votes" => 5
+      )
+      expect(Controller.new(double, action, params).expected_attributes(post).to_h).to eq("votes" => 5)
+    end
+
+    it "checks policy for expected attributes for record of a ActiveModel type" do
+      customer_post = Customer::Post.new(user)
+      params = to_params(
+        customer_post: {
+          title: "Hello",
+          votes: 5,
+          admin: true
+        }
+      )
+
+      action = "update"
+
+      expect(Controller.new(user, action, params).expected_attributes(customer_post).to_h).to eq(
+        "title" => "Hello",
+        "votes" => 5
+      )
+      expect(Controller.new(double, action, params).expected_attributes(customer_post).to_h).to eq(
+        "votes" => 5
+      )
+    end
+
+    it "goes through the policy cache" do
+      params = to_params(post: { title: "Hello" })
+      user = double
+      post = Post.new(user)
+      controller = Controller.new(user, "update", params)
+
+      expect do
+        expect(controller.expected_attributes(post)).to be_truthy
+        expect(controller.expected_attributes(post)).to be_truthy
+      end.to change { PostPolicy.instances }.by(1)
+    end
+  end
+
+  describe "#expected_attributes_for_action" do
+    it "is checked if it is defined in the policy" do
+      params = to_params(
+        post: {
+          title: "Hello",
+          body: "blah",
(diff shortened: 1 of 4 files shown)


spec/authorization_spec.rb
```diff
@@ -273,6 +273,91 @@ def to_params(*args, **kwargs, &block)
     end
   end
 
+  if Gem::Version.new(ActionPack::VERSION::STRING) >= Gem::Version.new("8.0.0")
```
Linuus, line 276: Looks like JRUBY doesn't support this version? I don't know if this is a good approach though :)

Burgestrand:
This looks as one could expect (:D). I'm also wondering if this is a good time to adjust what kind of API we will permit (:D), since this is an addition we've got time to adjust the API without breaking any backwards-compatibility.

Things I'm thinking of:
- what if our policy API is `expected_params_for(action)`, let the receiver worry about different params for different actions?
- param key, is it worth moving it to `pundit.param_key(record)` so it delegates to the proper policy finder

We're still rather light on Pundit-usage in our current projects. What about your project, do you use the action-specific `expected_params`?

jensljungblad-kisi:
> What about your project, do you use the action-specific expected_params?

We do, yeah.

Linuus:
> > What about your project, do you use the action-specific expected_params?
> 
> We do, yeah.

Well we are not using the expect at all yet as we are waiting for it to be added in Pundit, right? :)

jensljungblad-kisi:
Correct, I thought the question was if we're using the existing action specific permitted attributes methods.

Burgestrand:
Ah yes. Sometimes I write a bit för fort!

I haven't forgotten this. Still thinking about if I want to adjust the future API now that there's a chance to do so.

matthew-puku:
We use `permitted_attributes` 41 times and `permitted_attributes_for_x` 3 times.



> * what if our policy API is `expected_params_for(action)`, let the receiver worry about different params for different actions?

For us that means that in most cases the methods would look like:

```ruby
def expected_params_for(_action)
  %i[foo bar]
end
```

The action-specific cases would take one of two forms:

```ruby
# 1 Switch
def expected_params_for(action)
  case action.to_sym
  when :create
    %i[foo bar]
  when :update
    [:bar, (:foo if user.admin?)].compact
  else
    %i[bar]
  end
end

# 2 Concat
def expected_params_for(action)
  params = %i[foo]
  params += %i[bar baz] if action == :create
  params
end
```

Things I'd note:
- It's not immediately clear to me whether `action` should be a symbol or a string.
- We try to avoid custom actions, so for our project the `action` argument is essentially  a bool.
- The methods get gnarly if you have conditions on `action` and conditions on other stuff (as we usually do). 

---

All that to say, I _think_ I prefer copying the existing policy API ala `expected_attributes_for_x`, but it's not a huge deal either way.

Burgestrand:
@matthew-puku Thanks, this is _great_ feedback!

spec/authorization_spec.rb
Burgestrand: It's either this or checking for if the specific API method exists. JRuby 9.3 is EOL anyway: https://github.com/jruby/jruby/wiki/Roadmap

Burgestrand:
FYI I ended up rebasing this branch on `main`, rather than merging it into this one. Feature-wise I think we should be good now, and hopefully all checks pass now that I've updated the GH Actions matrix.

I'll let this marinate while I go to lunch, but I intend to merge it today.

I would be interested to hear your experiences if you're willing to try out running pundit from the main branch, to test expected attributes out.

Burgestrand:
Thank you! While this has been merged, I'm more than happy to keep options for improving this open. Once it's finally released the API should remain stable, but until then it's a free for all.

PR #858, opened 2025-05-06
Update inline code documentation

Burgestrand:
- **Move constants out of pundit.rb into conventional location**
- **Add version tags to each constant and method**

## To do


Burgestrand:
There should be no functional difference here, so the reasoning is we don't need a changelog.

PR #879, opened 2026-05-22
feat: add rubocop configuration for `Rails/ActionOrder`

G-Rath:
I realized this is a thing we can do, which I think is pretty cool 😄

## To do

- [x] ~I have added relevant tests.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add support for `params.expect` using `expected_parameters` and `expected_parameters_for`. [#855](https://github.com/varvet/pundit/pull/855)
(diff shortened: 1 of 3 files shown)


Burgestrand:
Hi! Thanks, yes, I agree this is indeed pretty cool!

I won't merge it, though. It's outside the scope of pundit to recommend you sort your controller actions in any specific way. At Varvet we usually use [standard-rails which has this rule disabled entirely](https://github.com/standardrb/standard-rails/blob/main/config/base.yml#L48-L49).

I think it would fit in our discussions under Show and Tell, though. I'm sure others would appreciate this nugget!
https://github.com/varvet/pundit/discussions/categories/show-and-tell

PR #881, opened 2026-08-03
fix #878: avoid policies action collisions

### LINKED ISSUE #878: Controller action named "policies" breaks everything
**Describe the bug**
Having a controller action with the name "policies" throws a very confusing error on all pages.

**To Reproduce**

1. Make a controller method named "policies"
2. Try to call "authorize(:something, something?)" in any action in that same controller

**Expected behavior**
I expect to be able to make a controller action named "policies" or at least display an error such as "invalid controller action named 'policies' "  Maybe just rename the internal "policies" method to "pundit_policies" to make it more likely to avoid namespace conflicts.

**Additional context**
I was on the lookout for this type of error, because I am making an app to store all of our "policies" :)  It is funny to have a PolicyPolicy :) This is the only issue I have come across so far. Thank you.

aouxwoux:
## motivation / background

Pundit::Authorization#pundit used policies as its cache hook. a controller action with the same name overrides that method, so its return value is passed to the policy cache and authorization fails.

fixes #878

## detail

this uses pundit_policies internally for the cache and keeps policies as a compatibility delegate. it also adds a regression spec for a controller with a policies action.

## checks

- [x] i have read the contributing guidelines
- [x] this pull request is related to one change
- [x] i have added relevant tests
- [x] i have adjusted relevant documentation
- [x] i have made sure the individual commit is meaningful
- [x] i have added a relevant changelog entry
- [x] bundle exec rspec: 162 examples, 0 failures
- [x] bundle exec standardrb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixed
 - Update for rspec 4 breaking changes [#873](https://github.com/varvet/pundit/issues/873)
+- Allow controllers to define a `policies` action without interfering with Pundit's policy cache. [#878](https://github.com/varvet/pundit/issues/878)
 
 ## 2.5.2 (2025-09-24)
 
--- a/lib/pundit/authorization.rb
+++ b/lib/pundit/authorization.rb
@@ -35,7 +35,7 @@ module Pundit
     def pundit
       @pundit ||= Pundit::Context.new(
         user: pundit_user,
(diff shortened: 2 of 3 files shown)


Burgestrand:
Hi! Thanks a lot for the PR, but there's already one waiting to be merged that I just haven't gotten around to yet 😅 

https://github.com/varvet/pundit/commit/5e62b1b49b72fda010fa154fbfc944509828e560

